Issue #1, opened 2020-02-06
make failes on error: cannot install, GOBIN must be an absolute path


RamLavi:
hack/install-go.sh 1.12.15
++ grep '^go ' go.mod
++ awk '{print $2}'
+ version=1.12.15
+ unset GOFLAGS
+ go get golang.org/dl/go1.12.15
go: finding golang.org/dl/go1.12.15 latest
go: finding golang.org/dl latest
cannot install, GOBIN must be an absolute path
make: *** [Makefile:15: /home/ralavi/go/bin/go1.12.15] Error 1

qinqon:
Thi sis my crazy approach, I am going for knmstate approach of just using tarball

qinqon:
I have re-wrote it, I am going to close the issue.

PR #2, opened 2020-02-06
add api to modify server outside of New method

RamLavi:
Currently you can modify server modifiers from New() method. 
It is in some cases required to add a server modifier outside of this method.
This PR exposes ServerUpdate() api method, to do just that.

### DIFF
--- a/pkg/webhook/server/server.go
+++ b/pkg/webhook/server/server.go
@@ -12,7 +12,7 @@ import (
 	certificate "github.com/qinqon/kube-admission-webhook/pkg/webhook/server/certificate"
 )
 
-type server struct {
+type Server struct {
 	mgr           manager.Manager
 	webhookName   string
 	webhookType   certificate.WebhookType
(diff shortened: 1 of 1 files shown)


RamLavi:
@qinqon - since I couldn't do make ([opened an issue](https://github.com/qinqon/kube-admission-webhook/issues/1)), then I did 'go mod tidy' etc.. manually.
But may need to be extra careful in order not to spoil your library

pkg/webhook/server/server.go
```diff
@@ -59,7 +58,13 @@ func WithCertDir(certDir string) ServerModifier {
 	}
 }
 
-func (s *server) Start(stop <-chan struct{}) error {
+func ServerUpdate(s *Server, serverOpts []ServerModifier) {
```
qinqon: Better use an object receiver
func (s *Server) Server(serverOpts ...ServerModifier)
so we can do 
server.UpdateOptions(webhookserver.WithHook....)

pkg/webhook/server/server.go
```diff
@@ -59,7 +58,13 @@ func WithCertDir(certDir string) ServerModifier {
 	}
 }
 
-func (s *server) Start(stop <-chan struct{}) error {
+func (s *Server) ServerUpdate(serverOpts []ServerModifier) {
```
qinqon: Call this UpdateOpts since Server is implicit.

pkg/webhook/server/server.go
```diff
@@ -59,7 +58,13 @@ func WithCertDir(certDir string) ServerModifier {
 	}
 }
 
-func (s *server) Start(stop <-chan struct{}) error {
+func (s *Server) UpdateOpts(serverOpts []ServerModifier) {
```
phoracek: Can this be called after the server is started? Please document it.

Also, usually we'd use a single argument of variable length https://gobyexample.com/variadic-functions, that should be easier to use than initializing a list just to pass arguments

RamLavi: I believe once the manager is started these parameters cannot be updated (@qinqon - am I right?).

PR #9, opened 2020-04-16
Discover the cluster CA certificate configmap

qinqon:
It check if system is running under openshift with `kubectl get co openshift-apiserver` and
in that case it uses the openshift configmap/field for ca-cert otherwise the k8s one

openshift configmap and field:
```golang
key := types.NamespacedName{
            Namespace: "openshift-config",
            Name:      "initial-kube-apiserver-server-ca",
        }
field := "ca-bundle.crt"
```

k8s configmap and field:
```golang
key := types.NamespacedName{
	Namespace: "kube-system",
	Name:      "extension-apiserver-authentication",
}
field := "client-ca-file",
```

### DIFF
--- a/pkg/webhook/server/server.go
+++ b/pkg/webhook/server/server.go
@@ -23,8 +23,6 @@ type Server struct {
 	log              logr.Logger
 }
 
-type ServerModifier func(w *Server)
-
 // Add creates a new Conditions Mutating Webhook and adds it to the Manager. The Manager will set fields on the Webhook
 // and Start it when the Manager is Started.
 func New(mgr manager.Manager, webhookName string, webhookType certificate.WebhookType, serverOpts ...ServerModifier) *Server {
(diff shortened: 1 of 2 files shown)


pkg/webhook/server/server.go
```diff
@@ -35,47 +33,21 @@ func New(mgr manager.Manager, webhookName string, webhookType certificate.Webhoo
...
-		mgr:              mgr,
-		log:              logf.Log.WithName("webhook/server"),
+		mgr: mgr,
+		log: logf.Log.WithName("webhook/server"),
 	}
-	s.updateServerOpts(serverOpts...)
 
-	return s
-}
+	// Use k8s cacert configmap by default
```
phoracek, line 40: Should not this be auto by default?

pkg/webhook/server/server.go
```diff
@@ -35,47 +33,21 @@ func New(mgr manager.Manager, webhookName string, webhookType certificate.Webhoo
...
-		s.caConfigMapKey = key
-	}
+	return s
 }
 
-func WithCaConfigMapField(field string) ServerModifier {
-	return func(s *Server) {
-		s.caConfigMapField = field
-	}
+func NewWithAutoCACert(mgr manager.Manager, webhookName string, webhookType certificate.WebhookType, serverOpts ...ServerModifier) *Server {
```
phoracek, line 48: This seems redundant

PR #15, opened 2020-05-13
Manage secret's owner correctly

qinqon:
In case secret is new it will set the service as the owner, if it's an update it will
not touch it.

### DIFF
--- a/pkg/webhook/server/certificate/manager_test.go
+++ b/pkg/webhook/server/certificate/manager_test.go
@@ -103,7 +103,14 @@ func TestWaitForDeadlineAndRotate(t *testing.T) {
 		},
 	}
 
-	objs := []runtime.Object{mutatingWebhookConfiguration}
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "fooWebhook",
+			Name:      "fooWebhook",
+		},
+	}
+
+	objs := []runtime.Object{mutatingWebhookConfiguration, service}
 
 	client := fake.NewFakeClient(objs...)
 	certsDuration := time.Minute
--- a/pkg/webhook/server/certificate/secret.go
+++ b/pkg/webhook/server/certificate/secret.go
@@ -3,6 +3,8 @@ package certificate
 import (
 	"context"
 
+	"github.com/pkg/errors"
+
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


pkg/webhook/server/certificate/secret.go
```diff
@@ -12,33 +14,53 @@ import (
...
+		corev1.TLSCertKey:       triple.EncodeCertPEM(keyPair.Cert),
+		corev1.TLSPrivateKeyKey: triple.EncodePrivateKeyPEM(keyPair.Key),
+	}
+	secret.Type = corev1.SecretTypeTLS
+	return &secret
+}
+
+func (m *Manager) newTLSSecret(serviceKey types.NamespacedName, keyPair *triple.KeyPair) (*corev1.Secret, error) {
+	service := corev1.Service{}
+	err := m.get(serviceKey, &service)
```
RamLavi, line 28: please incorporate this line in the if statement:
`if err := m.get(serviceKey, &service); err != nil {`

pkg/webhook/server/certificate/secret.go
```diff
@@ -12,33 +14,53 @@ import (
...
-	return secret
+	return updateTLSSecret(secret, keyPair), nil
 }
 
 func (m *Manager) applyTLSSecret(service types.NamespacedName, keyPair *triple.KeyPair) error {
-	tlsSecret := tlsSecret(service, keyPair)
+	secret := corev1.Secret{}
 
-	err := m.get(service, &corev1.Secret{})
+	err := m.get(service, &secret)
```
RamLavi, line 51: ditto

pkg/webhook/server/certificate/secret.go
```diff
@@ -12,33 +14,53 @@ import (
...
+			tlsSecret, err := m.newTLSSecret(service, keyPair)
+			if err != nil {
+				return errors.Wrapf(err, "failed initailizing secret %s", service)
+			}
+			return m.client.Create(context.TODO(), tlsSecret)
 		} else {
 			return err
 		}
 	}
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
RamLavi, line 63: [just asking to understand) if there's a conflict, don't we need to do an additional get in order to refresh the conflicted object (in our case, secret/keypair)?

pkg/webhook/server/certificate/manager_test.go
```diff
@@ -103,7 +103,14 @@ func TestWaitForDeadlineAndRotate(t *testing.T) {
 		},
 	}
 
-	objs := []runtime.Object{mutatingWebhookConfiguration}
+	service := &corev1.Service{
```
RamLavi, line 106: I'm not sure I understand why we are inserting a service here. can you explain?

pkg/webhook/server/certificate/secret.go
qinqon: The rest of the project use different line, is no good to change that here.

pkg/webhook/server/certificate/secret.go
qinqon: Same here,Ii prefer to have one responsability per line.

pkg/webhook/server/certificate/secret.go
qinqon: Ohhh man you are soo right, good catch !

pkg/webhook/server/certificate/manager_test.go
qinqon: If cert-manager is the one creating the secret we have to be sure that it's deleted if not needed, best way is to set the service as the owner so whenever the service is removed the secret will be removed as well.

RamLavi: ack

PR #17, opened 2020-05-18
fix index out of range error when trying to access client-config arra…

RamLavi:
…y in empty webhook instance

in order to avoid accessing client-config (that holds the caBundle in the mutating/validating webhook configuration instance) in a case where such webhook section does not exist, we added a protection check.

### DIFF
--- a/pkg/webhook/server/certificate/configuration.go
+++ b/pkg/webhook/server/certificate/configuration.go
@@ -109,5 +109,11 @@ func (m *Manager) CABundle() ([]byte, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get %s webhook configuration %s", m.webhookType, m.webhookName)
 	}
-	return m.clientConfigList(webhook)[0].CABundle, nil
+
+	clientConfigList := m.clientConfigList(webhook)
+	if clientConfigList == nil || len(clientConfigList) == 0 {
+		return nil, errors.Wrapf(err, "failed to access CABundle clientConfigList is empty in %s webhook configuration %s", m.webhookType, m.webhookName)
+	}
+
+	return clientConfigList[0].CABundle, nil
 }


pkg/webhook/server/certificate/configuration.go
```diff
@@ -109,5 +109,10 @@ func (m *Manager) CABundle() ([]byte, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get %s webhook configuration %s", m.webhookType, m.webhookName)
 	}
-	return m.clientConfigList(webhook)[0].CABundle, nil
+
+	if clientConfigList := m.clientConfigList(webhook); len(clientConfigList) == 0 {
+		return nil, errors.Wrapf(err, "failed to get clientConfig in %s webhook configuration %s", m.webhookType, m.webhookName)
```
ormergi, line 114: Can you add  "CABundle from" to the error message?
`"failed to get CABundle from clientConfig in %s webhook configuration %s"`

pkg/webhook/server/certificate/configuration.go
```diff
@@ -109,5 +109,10 @@ func (m *Manager) CABundle() ([]byte, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get %s webhook configuration %s", m.webhookType, m.webhookName)
 	}
-	return m.clientConfigList(webhook)[0].CABundle, nil
+
+	if clientConfigList := m.clientConfigList(webhook); len(clientConfigList) == 0 {
```
qinqon, line 113: else + returns is a "no-no"

Something like this is more idiomatic with golang and with the rest of the code here.
```golang
clientConfigList := m.clientConfigList(webhook)
if len(clientConfigList) == 0 {
   return nil, errors.Wrapf(err, "failed to get clientConfig in %s webhook configuration %s", m.webhookType, m.webhookName)
}

return clientConfigList[0].CABundle, nil
```

pkg/webhook/server/certificate/manager.go
```diff
@@ -117,7 +117,7 @@ func (m *Manager) Start(stopCh <-chan struct{}) error {
 // of generating new one
 func (m *Manager) loadCACertFromCABundle() {
 	caBundle, err := m.CABundle()
-	if err != nil || len(caBundle) == 0 {
```
qinqon, line 120: This is not related and also `ParseCertsPEM` will fail if caBundle is empty,

pkg/webhook/server/certificate/configuration.go
```diff
@@ -109,5 +109,10 @@ func (m *Manager) CABundle() ([]byte, error) {
...
+		return nil, errors.Wrapf(err, "failed to get clientConfig in %s webhook configuration %s", m.webhookType, m.webhookName)
+	} else {
+		return clientConfigList[0].CABundle, nil
```
ormergi, line 116: Shouldn't we also check if CABundel is not nil?

pkg/webhook/server/certificate/configuration.go
RamLavi: ack

pkg/webhook/server/certificate/configuration.go
qinqon: Agree let's add that check to first == nil then len 0 with different error message so we can differenciate.

pkg/webhook/server/certificate/manager.go
```diff
@@ -117,7 +117,7 @@ func (m *Manager) Start(stopCh <-chan struct{}) error {
 // of generating new one
```
ormergi, line 117: Hey what will happen if cert-manager is restarted and current CA is dummy cert
will it trigger new cert creation?
@RamLavi @qinqon

pkg/webhook/server/certificate/manager.go
```diff
@@ -117,7 +117,7 @@ func (m *Manager) Start(stopCh <-chan struct{}) error {
...
 func (m *Manager) loadCACertFromCABundle() {
 	caBundle, err := m.CABundle()
-	if err != nil || len(caBundle) == 0 {
+	if err != nil {
```
ormergi, line 120: Hey what will happen if cert-manager is restarted and the current CA is dummy or empty cert,
will it trigger new cert creation?
@RamLavi @qinqon

pkg/webhook/server/certificate/manager.go
RamLavi: If I remove this line entirely then I think we could have an issue in ParseCertsPEM(), because there we ask `for len(caBundle) > 0 {` so we might fail on that

pkg/webhook/server/certificate/manager.go
RamLavi: I think so yes. is that a problem?

PR #20, opened 2020-05-19
Implement certificate manager as a controller

qinqon:
All the rotation logic is call now from the controller's Reconcile function
at normal flow certificates and keys are creatd and Reconcile is programmed
to re-enqueue on certificate deadline.

In case of some of the k8s resources involved has bein created/update/deleted
the Reconcile function verify them and rotate certificate if needed.


pkg/certificate/controller.go
```diff
@@ -0,0 +1,95 @@
...
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("certificate-controller", mgr, controller.Options{Reconciler: m})
+	if err != nil {
+		return errors.Wrap(err, "failed instanciating certificate controller")
+	}
+
+	// Watch only events for selected m.webhookName
+	onEventForThisWebhook := predicate.Funcs{
```
RamLavi, line 34: so, every change in related secret/{mutating,validating}WebhookConfiguration will bring up the reconcile method.
Is cert expiration included as an event of such a kind that will cause the Reconcile to operate? what event specifically?

pkg/certificate/secret.go
```diff
@@ -64,3 +65,50 @@ func (m *Manager) applyTLSSecret(service types.NamespacedName, keyPair *triple.K
...
+	if !found {
+		return errors.New("TLS certs not found")
+	}
+
+	certs, err := triple.ParseCertsPEM(certsPEM)
+	if err != nil {
+		return errors.Wrap(err, "failed parsing PEM TLS certs")
+	}
+
+	cas := x509.NewCertPool()
```
RamLavi, line 98: This part is already used in checkTLS(). can you put it in a function and it both?

pkg/certificate/controller.go
qinqon: So, expiration is handled by changing `RequeueAfter` at `Reconcile` return if something not related to expiration happends and Reconcile is call, reconcile verify that we are not after deadline and just reset `RequeueAfter` without `rotate`.

pkg/certificate/manager.go
```diff
@@ -88,88 +84,31 @@ func NewManager(
...
-		m.waitForDeadlineAndRotate()
-	}, time.Second, stopCh)
-
-	return nil
-}
-
-// In case the manager is restarted we have to load the current CA instead
-// of generating new one
-func (m *Manager) loadCACertFromCABundle() {
+func (m *Manager) firstCACertFromCABundle() (*x509.Certificate, error) {
```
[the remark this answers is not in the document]
qinqon: Yep, good suggestiong, I will use `get` since it's ready doing a kubectl get kind of thing, usually at golang `accessors` don't have the `get` prefix, but this is more than getting a field at memory so we good with `get`.

pkg/certificate/secret.go
qinqon: Yep, let me refactor both so we reuse some code.

pkg/certificate/controller.go
RamLavi: ack, cool!

pkg/certificate/manager.go
```diff
@@ -208,28 +140,75 @@ func (m *Manager) rotate() error {
...
+	totalDuration := float64(notAfter.Sub(certificate.NotBefore))
+	deadline := certificate.NotBefore.Add(jitteryDuration(totalDuration))
 
 	m.log.Info(fmt.Sprintf("Certificate expiration is %v, rotation deadline is %v", notAfter, deadline))
 	return deadline
 }
 
+func (m *Manager) nextElapsedToRotate() time.Duration {
+	deadline := m.nextRotationDeadline()
+	now := m.now()
```
RamLavi, line 186: suggestion: since it takes a little time to rotate, shouldn't we try to rotate some time in advance **before** the cert expires?
Could be easily changed by setting the `now` to x minutes before now, if i'm not mistaken.
what do you think?

qinqon: That's already done, this rotates 10/30 % of time before expiration with some random jitter.

pkg/certificate/manager.go
```diff
@@ -208,28 +140,75 @@ func (m *Manager) rotate() error {
...
+	elapsedToRotate := deadline.Sub(now)
+	m.log.Info(fmt.Sprintf("nextElapsedToRotate {now: %s, deadline: %s, elapsedToRotate: %s}", now, deadline, elapsedToRotate))
+	return elapsedToRotate
+}
+
+// checkTLS will verify that the caBundle and Secret are valid and can
```
RamLavi, line 192: comment typo: checkTLS-->verifyTLS

pkg/certificate/secret.go
```diff
@@ -64,3 +64,30 @@ func (m *Manager) applyTLSSecret(service types.NamespacedName, keyPair *triple.K
 		return m.client.Update(context.TODO(), updateTLSSecret(secret, keyPair))
 	})
 }
+
+// checkTLS will verify that the caBundle and Secret are valid and can
```
RamLavi, line 68: comment typo: checkTLS-->verifyTLSSecret

pkg/webhook/server/server.go
```diff
@@ -102,39 +103,19 @@ func (s *Server) checkTLS() error {
...
-	}
-
-	if _, err := certs[0].Verify(opts); err != nil {
-		return errors.Wrap(err, "failed to verify certificate")
+	err = triple.VerifyTLS(certPEM, key, caPEM)
+	if err != nil {
+		return errors.Wrapf(err, "failed verifying %s/%s", certsPath, keyPath)
 	}
 
 	return nil
```
RamLavi, line 121: since in the logs we only see the negative logs like `E0601 09:25:57.673216       1 server.go:128] failed verifying /etc/webhook/certs/tls.crt//etc/webhook/certs/tls.key: failed parsing PEM TLS key: data does not contain a valid RSA or ECDSA private key` I think we should also add a positive log so that the user/developer will understand that the secret is good now.
maybe inside `VerifyTLS()`? what do you think?

pkg/certificate/secret.go
```diff
@@ -0,0 +1,131 @@
...
+			Annotations: map[string]string{},
+		},
+	}
+
+	err := m.setSecretOwnership(secretKey, &secret)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed setting ownership to secret %s", secretKey)
+	}
+
+	return updateTLSSecret(secret, keyPair), nil
```
RamLavi, line 83: looks like if we get into `newTLSSecret` we perform `updateTLSSecret` twice (here and at the end of `applyTLSSecret`: [here](https://github.com/qinqon/kube-admission-webhook/pull/20/files#diff-6ae9efaf7b3c3c8432f8afde3ab8e2daR102)). Is this necessary?

pkg/certificate/secret.go
```diff
@@ -0,0 +1,131 @@
...
+				return m.client.Create(context.TODO(), tlsSecret)
+			} else {
+				return err
+			}
+		}
+		return m.client.Update(context.TODO(), updateTLSSecret(secret, keyPair))
+	})
+}
+
+// checkTLS will verify that the caBundle and Secret are valid and can
```
RamLavi, line 106: function name in comment is copied, needs to be: checkTLS-->verifyTLSSecret

pkg/certificate/triple/cert.go
```diff
@@ -154,8 +161,8 @@ func GenerateSelfSignedCertKey(host string, alternateIPs []net.IP, alternateDNS
 // <host>_<ip>-<ip>_<alternateDNS>-<alternateDNS>.key
 // Certs/keys not existing in that directory are created.
 func GenerateSelfSignedCertKeyWithFixtures(host string, alternateIPs []net.IP, alternateDNS []string, fixtureDirectory string) ([]byte, []byte, error) {
-	validFrom := time.Now().Add(-time.Hour) // valid an hour earlier to avoid flakes due to clock skew
-	maxAge := time.Hour * 24 * 365          // one year self-signed certs
+	validFrom := Now().Add(-time.Hour) // valid an hour earlier to avoid flakes due to clock skew
```
RamLavi, line 164: general question. by making the cert valid an hour earlier then "now" - couldn't it make it so that 2 certs are valid in the same time?

pkg/certificate/controller_test.go
```diff
@@ -0,0 +1,221 @@
...
+		Expect(obtainedSecret.Data).ToNot(BeEmpty(), "should contain a secret with TLS key/cert")
+
+		return TLS{
+			caBundle:          cliConfig.CABundle,
+			certificate:       obtainedSecret.Data[corev1.TLSCertKey],
+			privateKey:        obtainedSecret.Data[corev1.TLSPrivateKeyKey],
+			secretAnnotations: obtainedSecret.Annotations,
+		}
+	}
+	Context("when reconcile is called for the fist time", func() {
```
RamLavi, line 90: typo fist--> first

pkg/certificate/secret.go
```diff
@@ -0,0 +1,131 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+
+	"github.com/qinqon/kube-admission-webhook/pkg/certificate/triple"
+)
+
+const (
+	secretManagedAnnotatoinKey = "kubevirt.io/kube-admission-webhook"
```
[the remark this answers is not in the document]
qinqon: I was tempted to use even `qinqon.io` usually those annotations points to project like the ones at multus for example.

pkg/certificate/secret.go
qinqon: Good catch !!

qinqon: If we create we don't reach the end of applyTLSSecret since we return at 
```golang
return m.client.Create(context.TODO(), tlsSecret)
```

pkg/certificate/triple/cert.go
qinqon: I copy pasted it from kubevirt but we don't need this functions, removing them, thanks !

PR #21, opened 2020-05-19
introducing a script to force rotate the kube-admission caBundle and secrets

RamLavi:
Since we noticed there may be a race between pod and secrets when manager pod is controlled by an extarnal operator, we want to introduce a workaround script, that deletes caBundle from the webhook config instance and restarts the operating pod. This triggers re-rotation of caBundle and secret.

### DIFF
--- a/hack/force-cert-retation.sh
+++ b/hack/force-cert-retation.sh
@@ -1,31 +1,153 @@
 #!/usr/bin/env bash
 
-set -xe
+set -ex
+
+function usage {
+cat <<- EOF
+Usage example:
+  HANDLER_NAMESPACE=kubemacpool-system HANDLER_POD_LABEL=kubemacpool-leader=true WEBHOOK_CONFIG_NAME=kubemacpool-mutator WEBHOOK_SECRET_NAME=kubemacpool-service ./force-cert-rotation.sh
+
+Mandatory Arguments:
+  HANDLER_NAMESPACE     namespace of the pod operating the kube-admission's library.
+  HANDLER_POD_LABEL     specific label corresponding to the pod operating the kube-admission's library.
+  WEBHOOK_CONFIG_NAME   webhook config name.
+  WEBHOOK_SECRET_NAME   the name of the secret created by kube-admission.
+EOF
+}
+
+# input parameter parsing
+function parse_input_param {
+  while [[ ! -z "$1" ]]; do
+    case "$1" in
+      --help)
+          usage
+          exit
+          ;;
+      *)
+          break
+          ;;
+      esac
+      shift
+  done
+}
+
+# check mandatory env variables
+function check_input_env_var {
+  if [[ -z "${HANDLER_NAMESPACE}" || -z "${HANDLER_POD_LABEL}" || -z "${WEBHOOK_CONFIG_NAME}" || -z "${WEBHOOK_SECRET_NAME}" ]]; then
+    echo "Error: Not all env parameters are set."
+    usage
+    exit
+  fi
+}
+
+# get number of webhooks in webhook config instance
+function get_number_of_webhooks_in_webhook_config {
+  echo $(${KUBECTL} get ${WEBHOOK_TYPE} -n ${HANDLER_NAMESPACE} ${WEBHOOK_CONFIG_NAME} -o jsonpath={.webhooks[*].name} | wc -w)
+}
+
+# retrive data from secret data
+function get_secret_data {
+  echo "$(${KUBECTL} get secret -n ${HANDLER_NAMESPACE} ${WEBHOOK_SECRET_NAME} -o jsonpath={.data})"
+}
+
+# retreive caBundle from webhook config
+function get_ca_bundle {
+  echo "$(${KUBECTL} get ${WEBHOOK_TYPE} -n ${HANDLER_NAMESPACE} ${WEBHOOK_CONFIG_NAME} -o jsonpath={.webhooks[$1].clientConfig.caBundle})"
+}
+
+# get the components to compare later
+function get_components_prior_to_rotation {
+  OLD_CA_BUNDLE=$(get_ca_bundle 0) 2>/dev/null
+
+  OLD_SECRET_DATA=$(get_secret_data) 2>/dev/null
+}
+
+# patch webhook config to remove caBundle
+function remove_ca_bundle_from_webhook_config {
+  for i in $(seq 0 $(( $number_of_webhooks - 1 )) ); do
+    ${KUBECTL} patch ${WEBHOOK_TYPE} -n ${HANDLER_NAMESPACE} ${WEBHOOK_CONFIG_NAME} --type json -p "[{\"op\": \"remove\", \"path\": \"/webhooks/$i/clientConfig/caBundle\"}]" || true
+  done
+}
+
+function restart_handler_pods {
+  # restart pods to force rotation of certs
+  ${KUBECTL} delete pod -n ${HANDLER_NAMESPACE} -l ${HANDLER_POD_LABEL}
+}
+
+function eventually {
+    timeout=100
+    interval=10
+    cmd=$@
+    echo "Checking eventually $cmd"
+    while ! $cmd; do
+        sleep $interval
+        timeout=$(( $timeout - $interval ))
+        if [ $timeout -le 0 ]; then
+            return 1
+        fi
+    done
+}
+
+function is_secret_rotated {
+  NEW_SECRET_DATA=$(get_secret_data) 2>/dev/null
+
+  [ "${NEW_SECRET_DATA}" != "${OLD_SECRET_DATA}" ]
+}
+
+function is_ca_bundle_rotated {
+  for i in $(seq 0 $(( $number_of_webhooks - 1 )) ); do
+      CA_BUNDLE=$(get_ca_bundle $i) 2>/dev/null
+
+      if [[ -z "${CA_BUNDLE}" || "${CA_BUNDLE}" == ${OLD_CA_BUNDLE} ]]; then
+        return 1
+      fi
+  done
+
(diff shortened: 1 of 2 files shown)


hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
phoracek, line 1: Please add a reference to the README with an example.

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -xe
+
+KUBECTL=${KUBECTL:-kubectl}
+POD_NAMESPACE=${POD_NAMESPACE:-kubemacpool-system}
```
phoracek, line 6: Let's prevent people from using these defaults by setting them to empty values. People may get used to the fact that this script works well with totally unrelated KMP and be surprised when it stops working. It's better to leave it empty, add a check that verifies that the variable was set and README with an example on how to set them.

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
...
+POD_APP_LABEL=${POD_APP_LABEL:-kubemacpool-leader=true}
```
phoracek, line 7: ditto

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
...
+WEBHOOK_TYPE=${WEBHOOK_TYPE:-mutatingWebhookConfiguration}
```
phoracek, line 8: ditto

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
...
+WEBHOOK_NAME=${WEBHOOK_NAME:-kubemacpool-mutator}
```
phoracek, line 9: ditto

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
...
+TEMP_FOLDER=${TEMP_FOLDER:-_tmp}
```
phoracek, line 10: use `mktemp -d`

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
...
+
+WEBHOOK_CONFIG_YAML_DIR=$(pwd)/${TEMP_FOLDER}/
+WEBHOOK_YAML_FILE=${WEBHOOK_CONFIG_YAML_DIR}/${WEBHOOK_TYPE}_${WEBHOOK_NAME}.yaml
+
+mkdir -p ${WEBHOOK_CONFIG_YAML_DIR}
+
+# get webhook config in yaml format
+${KUBECTL} get ${WEBHOOK_TYPE} -n ${POD_NAMESPACE} ${WEBHOOK_NAME} -oyaml > ${WEBHOOK_YAML_FILE}
```
phoracek, line 18: You can remove caBundle using `kubectl` with JSON patch.

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,31 @@
...
+
+#remove the caBundle from the webhook config
+sed '/caBundle:/d' ${WEBHOOK_YAML_FILE} > ${WEBHOOK_YAML_FILE}.out
```
phoracek, line 21: This will not work when caBundle is set as a multiline string

hack/force-cert-retation.sh
qinqon: This is `POD_LABEL` if we name it `POD_APP_LABEL` can be confused with usual `app` label.

hack/force-cert-retation.sh
qinqon: Let's remove this attribute and do it at all webhookconfigurations mutating or validating.

hack/force-cert-retation.sh
qinqon: and the proper cleanup at signal 

```bash
tmpdir=$(mktemp -d)
trap "{ rm -f $tmpdir; }" EXIT SIGINT SIGTERM SIGSTOP
```

hack/force-cert-retation.sh
qinqon: Examples of using it:

https://kubernetes.io/docs/reference/kubectl/cheatsheet/#patching-resources

hack/force-cert-retation.sh
RamLavi: I'll add help section to the script and reference it on README.md

hack/force-cert-retation.sh
RamLavi: In order to do that, I need to know the specific location of the caBundle, so I either need to add the webhooks names as env var or try to iterate on them using shyaml (not sure they support. sed was simpler.. 
What do you think? is there another way?

hack/force-cert-retation.sh
RamLavi: I'm not sure what you mean. isn't the caBundle a very long string by default?

hack/force-cert-retation.sh
RamLavi: well let me try first before I say it hard :)

hack/force-cert-retation.sh
phoracek: Both of the following are valid yaml strings represented as a single line:

```
foo: very long line
```

```
foo: "very
long
line"
```

```
foo: >
  very
  long
  line
```

Not sure they will be usefull for BASE64, since they include spaces between words though. But the point is, using singleline sed replacement may not always work. But if we always create a single line in this library, then this all is irrelevant :)

hack/force-cert-retation.sh
qinqon: I think you can use ".." to do nested CABundle patch at least jsonpath use that.

phoracek: But we can have multiple webhooks, so `..` would give us only the first one, no?

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,113 @@
...
+          break
+          ;;
+      esac
+      shift
+  done
+}
+
+# retrive tls.crt from secret data
+function get_secret_data {
+  echo "$(${KUBECTL} get secret -n ${HANDLER_NAMESPACE} ${WEBHOOK_SECRET_NAME} -oyaml | shyaml get-value data)"
```
RamLavi, line 52: I used shyaml here. not sure if it's the right way to go. what do you think?

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,113 @@
...
+function input_env_var_check {
+  # check mandatory env variables
+  if [[ -z "${HANDLER_NAMESPACE}" || -z "${HANDLER_POD_LABEL}" || -z "${WEBHOOK_TYPE}" || -z "${WEBHOOK_CONFIG_NAME}" || -z "${WEBHOOK_SECRET_NAME}" ]]; then
+    echo "Error: Not all env parameters are set."
+    usage
+    exit
+  fi
+
+  # check KUBECTL validity
+  if [[ -z "$KUBECTL" || -z "$($KUBECTL version)" ]]; then
```
qinqon, line 28: This is going to be defined we have a defaultm, remove this check

hack/force-cert-retation.sh
RamLavi: I looked at the pem format RFC. couldn't find a line like "PEM format is one line string" - but I'm fairly sure it is. @qinqon - do you think sed is the way to go ?

hack/force-cert-retation.sh
qinqon, line 52: You can use jsonpath to get data no need for `shyaml`

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,113 @@
...
+    while ! $cmd; do
+        sleep $interval
+        timeout=$(( $timeout - $interval ))
+        if [ $timeout -le 0 ]; then
+            return 1
+        fi
+    done
+}
+
+function isSecretRotated {
```
qinqon, line 69: is_secret_rotated, let's be consistent, also pass OLD_SECRET_DATA as an argument, so it's clear were stuff comes frome.

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,113 @@
...
+    echo "secret rotated successfuly"
+}
+
+KUBECTL=${KUBECTL:-kubectl}
+
+input_param_parse
+
+input_env_var_check
+
+WEBHOOK_CONFIG_YAML_DIR=$(mktemp -d --tmpdir=$(pwd))/
```
qinqon, line 90: We are missing the cleanup, but if you use jsonpatch this is not needed at all.

hack/force-cert-retation.sh
phoracek: Unless shyaml is a part of a base Fedora/CentOS distribution, it is a bad idea to use it in a generic script. If it is really that hard to use only `kubectl` commands to handle this, I won't fight against `sed`. However, I think you can iterate all webhook configs with `-o jsonpath='{ ... }'`

hack/force-cert-retation.sh
```diff
@@ -0,0 +1,113 @@
...
+WEBHOOK_YAML_FILE=${WEBHOOK_CONFIG_YAML_DIR}/${WEBHOOK_TYPE}_${WEBHOOK_CONFIG_NAME}.yaml
+
+mkdir -p ${WEBHOOK_CONFIG_YAML_DIR}
```
qinqon, line 93: not needed

hack/force-cert-retation.sh
RamLavi: ack. shyaml is indeed not a good option as it requires seperate pip install.

PR #26, opened 2020-06-17
Move unit tests to ginkgo

qinqon:
This provides powerful matchers and similar feel like other projects.


Makefile
```diff
@@ -45,8 +45,8 @@ format: $(FMT)
 vet: $(GO)
 	$(GO) vet ./pkg/...
 
-test: $(GO) vet format
-	$(GO) test -timeout 2m -v ./pkg/...
+test: $(GO)
```
RamLavi, line 48: if you remove  `vet` and `format` from here then they are not used at all. maybe we should move them to another dependency so it would checked regularly?

Makefile
```diff
@@ -45,8 +45,8 @@ format: $(FMT)
...
+	$(GO) test ./pkg/... -timeout 2m -ginkgo.v -test.v
```
ormergi, line 49: Should we specify noColor?

pkg/webhook/server/certificate/manager_test.go
```diff
@@ -1,184 +1,185 @@
...
-		t.Error("CA bundle not updated")
-	}
+			service = &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "fooWebhook",
+					Name:      "fooWebhook",
+				},
+			}
+		)
+		BeforeEach(func() {
```
RamLavi, line 122: can you add a check that there are no caBundles & secrets?

Makefile
qinqon: I am going to put them back and fix this at a different PR, so we do similar that we do with KMP running them at CI and checking that they don't change code.

pkg/webhook/server/certificate/manager_test.go
```diff
@@ -1,184 +1,185 @@
...
+
+				err = client.Get(context.TODO(), types.NamespacedName{Name: "fooWebhook", Namespace: "fooWebhook"}, &secret)
+				Expect(err).ToNot(HaveOccurred(), "should success getting secret")
+
+				start = time.Now()
+				By("Calling waitForDeadlineAndRotate for the second time")
+				manager.waitForDeadlineAndRotate()
+			})
+
+			It("should wait some time, update the secret and re-rotate certificates", func() {
```
RamLavi, line 167: I don't understand something here.
we expect that certs should be rotated, but it is not clear why they've been rotated (expired?/secret deleted?/else?)
it should be clear from the Context what type of re-rotation we are testing.

RamLavi, line 167: a technical issue:
I assume we are checking re-rotation sue to expiration, but if so, then I don't understand where do we `wait()` for it to expire?

pkg/webhook/server/certificate/certificate_suite_test.go
```diff
@@ -0,0 +1,26 @@
...
+	. "github.com/onsi/gomega"
+
+	"github.com/onsi/ginkgo/reporters"
+
+	"k8s.io/klog"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+func init() {
+
```
RamLavi, line 16: can you remove the empty line?

pkg/webhook/server/server_suite_test.go
```diff
@@ -0,0 +1,26 @@
...
+	. "github.com/onsi/gomega"
+
+	"github.com/onsi/ginkgo/reporters"
+
+	"k8s.io/klog"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+func init() {
+
```
RamLavi, line 16: ditto

pkg/webhook/server/certificate/manager_test.go
qinqon: We are creating a new FakeClient so there should be none I suppose, right ?

RamLavi: yeah but if in the future there will be more then 1 test then we might end up with secrets from former tests.

qinqon: What I mean is that we store stuff at FakeClient memory and FakeClietn is re-created at BeforeEach so there is no issue

pkg/webhook/server/certificate/manager_test.go
qinqon: We wait at the `BeforeEach` with the `waitForDeadlineAndRotate` call.

pkg/webhook/server/certificate/manager_test.go
RamLavi: ack, thanks

pkg/webhook/server/server_suite_test.go
RamLavi: empty line not resolved

PR #28, opened 2020-07-07
Implement caBundle overlapping

qinqon:
Right now caBundle only store the rotated ca and discard the old ca that was previously there, this introduces problems since it takes some time for the secret to be updated into the pod when it's mounted, during this time if caBundle does not conserve the previous CA webhook traffic will fail.

To do that this PR add the rotate CA to ca bundle instead of replace it and implements at caBundle cleanup mechanism based on expiration time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@ The library generates RSA keys with 2048 size and certificate for both for CA an
 They share the expiration time so all the CA and service certificates
 are rotated at once just before expiration time.
 
+The CA bundle from webhook configuratin contains not only the last rotated
+CA certificate but also the non expired previous one, that prevents problems
+related to pods watching an old projection of the mounted secret.
+
 ## Webhook service
 It has a one year expiration time harcoded and apart from wrapping the
 controller runtime webhook library it waits for TLS key/cert existence and
--- a/pkg/certificate/cleanup.go
+++ b/pkg/certificate/cleanup.go
@@ -1,0 +1,85 @@
+package certificate
+
+import (
+	"crypto/x509"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/qinqon/kube-admission-webhook/pkg/certificate/triple"
+)
+
+// earliestElapsedForCleanup return a subtraction between earliestCleanupDeadline and
+// `now`
+func (m *Manager) earliestElapsedForCleanup() (time.Duration, error) {
+	deadline, err := m.earliestCleanupDeadline()
+	if err != nil {
+		return time.Duration(0), errors.Wrap(err, "failed calculating cleanup deadline")
+	}
+	now := m.now()
+	elapsedForCleanup := deadline.Sub(now)
+	m.log.Info(fmt.Sprintf("earliestElapsedForCleanup {now: %s, deadline: %s, elapsedForCleanup: %s}", now, deadline, elapsedForCleanup))
+	return elapsedForCleanup, nil
+}
+
+// earliestCleanupDeadline get all the certificates at CABundle and return the
+// deadline calculated by earliestCleanupDeadlineForCerts
+func (m *Manager) earliestCleanupDeadline() (time.Time, error) {
+	cas, err := m.getCACertsFromCABundle()
+	if err != nil {
+		return m.now(), errors.Wrap(err, "failed getting CA certificates from CA bundle")
+	}
+
+	return m.earliestCleanupDeadlineForCerts(cas), nil
+}
+
+// earliestCleanupDeadlineForCACerts will inspect CA certificates
+// select the deadline based on certificate that is going to expire
+// sooner so cleanup is triggered then
+func (m *Manager) earliestCleanupDeadlineForCerts(certificates []*x509.Certificate) time.Time {
+
+	var selectedCertificate *x509.Certificate
+
+	for _, certificate := range certificates {
+		if selectedCertificate == nil || certificate.NotAfter.Before(selectedCertificate.NotAfter) {
+			selectedCertificate = certificate
+		}
+	}
+	if selectedCertificate == nil {
+		return m.now()
+	}
+
+	return selectedCertificate.NotAfter
+}
+
+func (m *Manager) cleanUpCABundle() error {
+	logger := m.log.WithName("cleanUpCABundle")
+	logger.Info("Cleaning up expired certificates at CA bundle")
+	_, err := m.updateWebhookCABundleWithFunc(func([]byte) ([]byte, error) {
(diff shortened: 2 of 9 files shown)


pkg/certificate/cleanup.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed updating webhook config after ca certificates cleanup")
+	}
+	return nil
+}
+
+func (m *Manager) cleanUpExpiredCertificates(certificates []*x509.Certificate) []*x509.Certificate {
+	now := m.now()
+	// create a zero-length slice with the same underlying array
+	cleanedUpCertificates := certificates[:0]
```
RamLavi, line 78: just a golang implementation question. why are we using the same array instead of adding to a new array of certsToDelete?

pkg/certificate/controller.go
```diff
@@ -97,10 +109,40 @@ func (m *Manager) Reconcile(request reconcile.Request) (reconcile.Result, error)
...
+	}
+
+	elapsedForCleanup, err := m.nextElapsedForCleanup()
+	if err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "failed getting cleanup deadline")
+	}
+
+	// We have pass cleanup deadline let's do the cleanup
+	if elapsedForCleanup <= 0 {
+
```
RamLavi, line 124: please remove empty line

pkg/certificate/controller.go
```diff
@@ -97,10 +109,40 @@ func (m *Manager) Reconcile(request reconcile.Request) (reconcile.Result, error)
...
+		err = m.cleanUpCABundle()
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "failed cleaning up CABundle")
+		}
+
+		// Re-calculate cleanup deadline since we may have have remove some certs there
```
RamLavi, line 130: typo: have have-->have to

pkg/certificate/cleanup.go
qinqon: To re-use memory

pkg/certificate/manager.go
```diff
@@ -178,13 +188,18 @@ func (m *Manager) nextRotationDeadline() time.Time {
 		return m.now()
 	}
 
-	caCert, err := m.getFirstCACertFromCABundle()
+	// Last rotated CA cert at CABundle is the last at the slice so this
+	// calculate deadline from it.
+	caCert, err := m.getLastCACertFromCABundle()
 	if err != nil {
 		m.log.Info("Failed reading first CA cert from CABundle, forcing rotation", "err", err)
```
RamLavi, line 195: typo: `Failed reading first`-->`Failed reading last`

pkg/certificate/manager.go
```diff
@@ -178,13 +188,18 @@ func (m *Manager) nextRotationDeadline() time.Time {
 		return m.now()
 	}
 
-	caCert, err := m.getFirstCACertFromCABundle()
+	// Last rotated CA cert at CABundle is the last at the slice so this
+	// calculate deadline from it.
+	caCert, err := m.getLastCACertFromCABundle()
```
RamLavi, line 193: I still don't get why we take the last. 
my reasoning is: last caBundle is the last rotated, so it's expiration will be the furthest. don't we want the one with the earliest expiration?

pkg/certificate/cleanup.go
```diff
@@ -0,0 +1,85 @@
...
+	for _, certificate := range certificates {
+		if certificate.NotAfter != m.now() && certificate.NotAfter.After(now) {
```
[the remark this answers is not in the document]
qinqon: Now that I see it's impossible to have NotAfter equals to now at After now so the comparation with != does not make sense.

pkg/certificate/cleanup_test.go
```diff
@@ -0,0 +1,301 @@
...
+			})
+		}
+		return certificates
+	}
+
+	type nextCleanupDeadlineCase struct {
+		certsExpiration []certificateExpiration
+		expectedElapsed time.Duration
+	}
+	DescribeTable("NextCleanupCABundle",
```
RamLavi, line 38: I think you mean nextCleanupDeadlineForCerts right?

pkg/certificate/configuration.go
```diff
@@ -78,10 +78,29 @@ func (m *Manager) readyWebhookConfiguration() (runtime.Object, error) {
...
+	webhook, err := m.updateWebhookCABundleWithFunc(func(currentCABundle []byte) ([]byte, error) {
+		cas := []*x509.Certificate{}
+		if len(currentCABundle) > 0 {
+			var err error
+			cas, err = triple.ParseCertsPEM(currentCABundle)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed parsing current CA bundle")
+			}
+		}
+		cas = append(cas, caCert)
```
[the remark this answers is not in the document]
qinqon: We can implement this as a set using a golang map but I think it does not worth comparing a big pile of bytes.

pkg/certificate/manager.go
qinqon: This is the rotation deadline not the cleanup deadline.

RamLavi: ok, but why the last?

qinqon: Because we `append` the rotated cert so the last certificate is the one rotated.

RamLavi: ok I understand. 
Then since you are relying on the functionality of `append` function, maybe consider to change the function name to `getLastAppendedCACertFromCABundle`. I know it's long but I personally prefer it to be more readable like so.

RamLavi:
@qinqon as I see it, the main way you implemented the overlapping is to separate the `rotate()` method that only adds new certs to caBundles, from `cleanUpCABundle()` that removes expires caBundle. 
I like it, however as it's a subtle yet important change I think we should make it more clear by mentioning it in the commit documentation, and perhaps also in the README.

qinqon:
@RamLavi I will add it to the README, squash the commits and add a proper commit message.

README.md
```diff
@@ -6,6 +6,10 @@ The library generates RSA keys with 2048 size and certificate for both for CA an
 They share the expiration time so all the CA and service certificates
 are rotated at once just before expiration time.
 
+The CA bundle from webhook configuratin contains not only the last rotated
+CA certificate but also the non expired previous one, that prevents problems
+related to pods watching and old projection of the mounted secret.
```
RamLavi, line 11: typo "watching and old.."--> "watching an old.."

PR #29, opened 2020-07-10
Add  CARotationInverval knob

qinqon:
It will define the ca certificate and server certificate expiration time.

### DIFF
--- a/pkg/webhook/server/server.go
+++ b/pkg/webhook/server/server.go
@@ -33,13 +33,13 @@ type ServerModifier func(w *webhook.Server)
 
 // Add creates a new Conditions Mutating Webhook and adds it to the Manager. The Manager will set fields on the Webhook
 // and Start it when the Manager is Started.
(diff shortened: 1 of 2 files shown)


pkg/webhook/server/server.go
```diff
@@ -33,13 +33,13 @@ type ServerModifier func(w *webhook.Server)
 
 // Add creates a new Conditions Mutating Webhook and adds it to the Manager. The Manager will set fields on the Webhook
 // and Start it when the Manager is Started.
-func New(client client.Client, webhookName string, webhookType certificate.WebhookType, serverOpts ...ServerModifier) *Server {
+func New(client client.Client, webhookName string, webhookType certificate.WebhookType, caRotateInterval time.Duration, serverOpts ...ServerModifier) *Server {
```
phoracek, line 36: Should not we use ServerModifier for that?

qinqon: we don't want to make it optional.

PR #31, opened 2020-07-16
Skip Reconcile at Secret cascade delete

qinqon:
A cascade delete of a Secret means that the Services or owner is gone so secret should
be gone too, if we don't skip Reconcile there the secret could be re-created if Service
is not gone at that time.

### DIFF
--- a/pkg/certificate/controller.go
+++ b/pkg/certificate/controller.go
@@ -1,6 +1,7 @@
 package certificate
 
 import (
+	"context"
 	"fmt"
 	"time"
 
@@ -8,7 +9,11 @@ import (
 
 	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	unstructured "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -33,9 +38,38 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return errors.Wrap(err, "failed instanciating certificate controller")
 	}
 
-	isWebhookConfigOrAnnotatedResource := func(meta metav1.Object) bool {
+	isACascadeDelete := func(deleteEvent event.DeleteEvent) bool {
+		namespace := deleteEvent.Meta.GetNamespace()
+		for _, ownerRefernce := range deleteEvent.Meta.GetOwnerReferences() {
+			gv, err := schema.ParseGroupVersion(ownerRefernce.APIVersion)
+			if err != nil {
+				logger.Error(err, "failed parsing gv from owner")
+				continue
+			}
+			gvk := gv.WithKind(ownerRefernce.Kind)
+			object := unstructured.Unstructured{}
+			object.SetGroupVersionKind(gvk)
+			err = m.client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: ownerRefernce.Name}, &object)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					logger.Info("Cascade delete detected, at least one owner is not present")
+					return true
+				}
+				logger.Error(err, "failed detecting if this is a cascade delete")
+				continue
+			}
+		}
+		logger.Info("Cascade delete not-detected, all the owners are present")
+		return false
+	}
+
+	isAnnotatedResource := func(meta metav1.Object) bool {
 		_, foundAnnotation := meta.GetAnnotations()[secretManagedAnnotatoinKey]
-		return meta.GetName() == m.webhookName || foundAnnotation
+		return foundAnnotation
+	}
+
+	isWebhookConfigOrAnnotatedResource := func(meta metav1.Object) bool {
+		return meta.GetName() == m.webhookName || isAnnotatedResource(meta)
 	}
 
 	// Watch only events for selected m.webhookName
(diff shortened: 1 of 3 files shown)


pkg/certificate/controller.go
```diff
@@ -33,9 +38,38 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return errors.Wrap(err, "failed instanciating certificate controller")
 	}
 
-	isWebhookConfigOrAnnotatedResource := func(meta metav1.Object) bool {
+	isACascadeDelete := func(deleteEvent event.DeleteEvent) bool {
```
phoracek, line 41: Please use real functions.

pkg/certificate/controller.go
```diff
@@ -33,9 +38,38 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		return false
+	}
+
+	isAnnotatedResource := func(meta metav1.Object) bool {
 		_, foundAnnotation := meta.GetAnnotations()[secretManagedAnnotatoinKey]
-		return meta.GetName() == m.webhookName || foundAnnotation
+		return foundAnnotation
+	}
+
+	isWebhookConfigOrAnnotatedResource := func(meta metav1.Object) bool {
```
phoracek, line 71: This function alone does not bring any value. Better to have isAnnotatedResource and isWebhookConfig and then call them both with || in between.

pkg/certificate/controller.go
```diff
@@ -33,9 +38,38 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+				logger.Error(err, "failed parsing gv from owner")
+				continue
+			}
+			gvk := gv.WithKind(ownerRefernce.Kind)
+			object := unstructured.Unstructured{}
+			object.SetGroupVersionKind(gvk)
+			err = m.client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: ownerRefernce.Name}, &object)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					logger.Info("Cascade delete detected, at least one owner is not present")
```
phoracek, line 55: Is not present where? Please make it more verbose

pkg/certificate/controller.go
```diff
@@ -44,7 +78,8 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
 			return isWebhookConfigOrAnnotatedResource(createEvent.Meta)
 		},
 		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
-			return isWebhookConfigOrAnnotatedResource(deleteEvent.Meta)
+			_, isASecret := deleteEvent.Object.(*corev1.Secret)
```
phoracek, line 81: The sematics would be clearer if we have a separate branch for secret, so something like:

```
if isASecret {
    return special logic
}
return false
```

That way it is clear that for non-secrets it is always false without reading into the logic expression

pkg/certificate/controller.go
phoracek, line 55: Maybe it would be easier to understand if we called it "garbage collection detected"? It is more known than "cascade"

phoracek:
Please make the commit message more detailed. I want to know why are we changing this in the first place, what was the bug. I want to know what is the suggested fix and why we decided to do it this specific way.

Next week we will scratch our heads asking what did we mean by this. Let's make it easier.

pkg/certificate/controller.go
phoracek: And add at least a comment explaining that we do not want to recreate Secrets on deletion of the whole app. We want to recreate it only when they were deleted alone, by mistake or purpose to recreate a new cert.

qinqon:
KMP -> https://github.com/k8snetworkplumbingwg/kubemacpool/pull/232
CNAO -> https://github.com/kubevirt/cluster-network-addons-operator/pull/468

pkg/certificate/controller.go
```diff
@@ -44,7 +78,8 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			return isASecret && isAnnotatedResource(deleteEvent.Meta) && !isACascadeDelete(deleteEvent)
```
phoracek, line 82: This controller watches webhook configurations too. We don't care about those on deletion? That seems to be a change compared to the original code.

pkg/certificate/controller.go
```diff
@@ -33,9 +38,38 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		for _, ownerRefernce := range deleteEvent.Meta.GetOwnerReferences() {
+			gv, err := schema.ParseGroupVersion(ownerRefernce.APIVersion)
+			if err != nil {
+				logger.Error(err, "failed parsing gv from owner")
+				continue
+			}
+			gvk := gv.WithKind(ownerRefernce.Kind)
+			object := unstructured.Unstructured{}
+			object.SetGroupVersionKind(gvk)
+			err = m.client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: ownerRefernce.Name}, &object)
```
phoracek, line 52: I know we can probably expect that we are always in the same namespace but that is implicit. Could we use ownerReference.Namespace instead?

pkg/certificate/controller.go
```diff
@@ -33,9 +38,38 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			if err != nil {
+				if apierrors.IsNotFound(err) {
```
phoracek, line 54: The deletion is not sequential and blocking. We should check for the deletion timestamp in case we find it. But even then, we can have a race and it can also be caused by somebody deleting both at the same time, not GC.

If we use blocking garbage collection, can we then read the annotation you showed me describing that "this is being deleted as  a part of GC"?

pkg/certificate/controller_test.go
```diff
@@ -223,6 +226,31 @@ var _ = Describe("Certificates controller", func() {
...
+				err := cli.Get(context.TODO(), types.NamespacedName{Namespace: expectedSecret.Namespace, Name: expectedSecret.Name}, &obtainedSecret)
+				Expect(err).To(Succeed(), "should succeed getting the secret")
+				obtainedSecret.OwnerReferences[0].Name = "bad-service"
+				err = cli.Update(context.TODO(), &obtainedSecret)
+				Expect(err).To(Succeed(), "should succeed updating the TLS secret")
+
+				By("Wait a little for Reconcile to settle after updating secret")
+				time.Sleep(3 * time.Second)
+
+				By("Delete the TLS secret")
```
RamLavi, line 244: can you explain how does this mimic cascade delete of the secret? cascase is when the object is delete indirectly by deleting its owner, no?

PR #34, opened 2020-07-20
Use new knob for the service certificate rotation

qinqon:
This PR add a new "knob" `certRotationInterval` to specify the service certificate expiration time, the current version at master use the value at `CARotationInterval`. 

To accomplish this new secret is created with key `[namespace]/[webhookname]-ca`  the `namespace` and `webhookname` are the one passed to the webhook server.

Also a bug is fixed, the service secret was being re-constructed per webhook since it was just iterating instead of listing services and create new secret per each one.

### DIFF
--- a/pkg/certificate/configuration.go
+++ b/pkg/certificate/configuration.go
@@ -151,6 +151,7 @@ func (m *Manager) CABundle() ([]byte, error) {
 // passing the url hostname at map value
 func (m *Manager) getServicesFromConfiguration(configuration runtime.Object) (map[types.NamespacedName][]string, error) {
 
+	logger := m.log.WithName("getServicesFromConfiguration")
 	services := map[types.NamespacedName][]string{}
 
 	for _, clientConfig := range m.clientConfigList(configuration) {
@@ -159,9 +160,11 @@ func (m *Manager) getServicesFromConfiguration(configuration runtime.Object) (ma
 		hostnames := []string{}
 
 		if clientConfig.Service != nil {
+			logger.Info("Composing service name and namespace from ServiceRef", "serviceRef", clientConfig.Service)
 			service.Name = clientConfig.Service.Name
 			service.Namespace = clientConfig.Service.Namespace
 		} else if clientConfig.URL != nil {
+			logger.Info("Composing service name and namespace from URL", "URL", clientConfig.URL)
 			service.Name = m.webhookName
 			service.Namespace = "default"
 			u, err := url.Parse(*clientConfig.URL)
@@ -169,6 +172,8 @@ func (m *Manager) getServicesFromConfiguration(configuration runtime.Object) (ma
 				return nil, errors.Wrapf(err, "failed parsing webhook URL %s", *clientConfig.URL)
 			}
 			hostnames = append(hostnames, strings.Split(u.Host, ":")[0])
+		} else {
+			return nil, errors.New("bad configuration, webhook without serviceRef or URL")
 		}
 
 		services[service] = hostnames
--- a/pkg/certificate/configuration.go
+++ b/pkg/certificate/configuration.go
@@ -147,7 +147,7 @@ func (m *Manager) CABundle() ([]byte, error) {
 
 // getServicesFromConfiguration it retrieves all the references services at
 // webhook configuration clientConfig and in case there is no service ref
-// it will refernece fake one with webhook name, default namespaces and
+// it will refernece fake one with webhook name, mgr namespace and
 // passing the url hostname at map value
 func (m *Manager) getServicesFromConfiguration(configuration runtime.Object) (map[types.NamespacedName][]string, error) {
 
@@ -166,7 +166,7 @@ func (m *Manager) getServicesFromConfiguration(configuration runtime.Object) (ma
 		} else if clientConfig.URL != nil {
 			logger.Info("Composing service name and namespace from URL", "URL", clientConfig.URL)
 			service.Name = m.webhookName
-			service.Namespace = "default"
+			service.Namespace = m.namespace
 			u, err := url.Parse(*clientConfig.URL)
 			if err != nil {
 				return nil, errors.Wrapf(err, "failed parsing webhook URL %s", *clientConfig.URL)
--- a/pkg/certificate/configuration.go
+++ b/pkg/certificate/configuration.go
@@ -80,9 +80,9 @@ func (m *Manager) readyWebhookConfiguration() (runtime.Object, error) {
 	return webhook, err
 }
 
-func (m *Manager) addCertificateToCABundle(caCert *x509.Certificate) (runtime.Object, error) {
+func (m *Manager) addCertificateToCABundle(caCert *x509.Certificate) error {
 	m.log.Info("Reset CA bundle with one cert for webhook")
-	webhook, err := m.updateWebhookCABundleWithFunc(func(currentCABundle []byte) ([]byte, error) {
+	_, err := m.updateWebhookCABundleWithFunc(func(currentCABundle []byte) ([]byte, error) {
 		cas := []*x509.Certificate{}
 		if len(currentCABundle) > 0 {
 			var err error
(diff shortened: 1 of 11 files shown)


pkg/certificate/controller.go
```diff
@@ -89,6 +89,7 @@ func (m *Manager) Reconcile(request reconcile.Request) (reconcile.Result, error)
 	reqLogger.Info("Reconciling Certificates")
 
 	elapsedToRotate := m.elapsedToRotateFromLastDeadline()
```
phoracek, line 91: You added ToRotateServices. This is to rotate what?

pkg/certificate/secret.go
```diff
@@ -42,19 +42,18 @@ func populateTLSSecret(secret corev1.Secret, keyPair *triple.KeyPair) *corev1.Se
...
 }
 
 func (m *Manager) applyTLSSecret(secret types.NamespacedName, keyPair *triple.KeyPair) error {
-	return m.applySecret(secret, keyPair, populateTLSSecret)
+	return m.applySecret(secret, corev1.SecretTypeTLS, keyPair, populateTLSSecret)
 }
 
 func (m *Manager) applyCASecret(secret types.NamespacedName, keyPair *triple.KeyPair) error {
-	return m.applySecret(secret, keyPair, populateCASecret)
+	return m.applySecret(secret, corev1.SecretTypeOpaque, keyPair, populateCASecret)
```
RamLavi: isn't it strange that there is no `corev1.SecretTypeCA`?

pkg/certificate/secret.go
```diff
@@ -112,3 +112,33 @@ func (m *Manager) verifyTLSSecret(secretKey types.NamespacedName, caBundle []byt
 
 	return nil
 }
+
+func (m *Manager) getCAKeyPair() (*triple.KeyPair, error) {
+	secret := corev1.Secret{}
```
RamLavi, line 133: please change secret var name to caSecret for clarity

pkg/certificate/controller.go
RamLavi, line 91: please change to elapsedToRotateCAFromLastDeadline

pkg/certificate/controller.go
```diff
@@ -89,6 +89,7 @@ func (m *Manager) Reconcile(request reconcile.Request) (reconcile.Result, error)
...
+	elapsedToRotateServices := m.elapsedToRotateServicesFromLastDeadline()
 
 	// Ensure that this Reconcile is not called after bad changes at
 	// the certificate chain
```
RamLavi, line 95: why do we only consider `if elapsedToRotate > 0` to check TLS verify? shouldn't we also account for the elapsedToRotateServices?

pkg/certificate/manager.go
```diff
@@ -190,6 +191,42 @@ func (m *Manager) rotateServices() error {
...
+
+	services, err := m.getServicesFromConfiguration(webhookConf)
+	if err != nil {
+		m.log.Info(fmt.Sprintf("failed getting webhook configuration services, forcing rotation: %v", err))
+		return m.now()
+	}
+
+	// Iterate the `services` map to calculate deadline with the first
+	// occurrence
+	for service, _ := range services {
```
RamLavi, line 224: If I understand this correctly, then right now we run a for loop of the services, but we return the first one.
Shouldn't we return the one with the earliest deadline out of the services?

pkg/certificate/manager_test.go
```diff
@@ -116,7 +116,7 @@ var _ = Describe("certificate manager", func() {
 	loadCASecret := func(manager *Manager) corev1.Secret {
 		secretKey := types.NamespacedName{
 			Namespace: "default",
-			Name:      expectedMutatingWebhookConfiguration.Name,
+			Name:      expectedMutatingWebhookConfiguration.Name + "-ca",
```
RamLavi: can we use `caSecretKey()` here?

pkg/certificate/manager.go
```diff
@@ -123,38 +135,51 @@ func (m *Manager) getLastAppendedCACertFromCABundle() (*x509.Certificate, error)
...
+	m.log.Info("Rotating CA cert/key")
 
-	caKeyPair, err := triple.NewCA(m.webhookName, m.certsDuration)
+	caKeyPair, err := triple.NewCA(m.webhookName, m.caCertDuration)
 	if err != nil {
 		return errors.Wrap(err, "failed generating CA cert/key")
 	}
 
-	webhook, err := m.addCertificateToCABundle(caKeyPair.Cert)
+	err = m.addCertificateToCABundle(caKeyPair.Cert)
```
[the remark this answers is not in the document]
qinqon: Nope, this is the library adding the certificate to webhook configuration CABundle field.

pkg/certificate/configuration.go
```diff
@@ -142,3 +144,34 @@ func (m *Manager) CABundle() ([]byte, error) {
 
 	return clientConfigList[0].CABundle, nil
 }
+
+// getServicesFromConfiguration it retrieves all the references services at
+// webhook configuration clientConfig and in case there is no service ref
+// it will refernece fake one with webhook name, default namespaces and
```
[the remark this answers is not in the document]
qinqon: Fixed a little the comment, discovered this after running tests with envtest we cannot use Services there since it's not a proper cluster to exercise webhooks we use URL field instead of ServiceRef.

pkg/certificate/secret.go
qinqon: It's not, I though the same.

pkg/certificate/manager.go
qinqon: We create service certificates at the same time so they have almost the same deadline so no need for this.

pkg/certificate/manager_test.go
qinqon: We don't have the manager here so we can call `caSecretKey`

pkg/certificate/manager.go
RamLavi: yes but it's a private case behavior that might change in the future. I think we should keep it general, i.e., go over all services and take the earliest one.

qinqon: fair enough.

pkg/certificate/controller.go
```diff
@@ -101,21 +102,36 @@ func (m *Manager) Reconcile(request reconcile.Request) (reconcile.Result, error)
 		}
```
RamLavi, line 102: Regarding Reconsile, what happens if we delete one of the services?
For example, let's say there are 2 services, that were created when the server was created, with 1year expiration. After 0.5year,  one of the services are deleted, and recreated by the controller. 
1. Will we notice it in this reconcile?
2. if  we do, do we recreate all of the services at once?

qinqon: I don't know if you can remove a Services that is references by a webhook configuration, maybe kube-apiserver prevent it, anyhow Service will be re-created and secret still valid, I suppose the services will be reacreated by CNAO right ? we don't re-create them here.

pkg/certificate/controller.go
```diff
@@ -88,7 +88,8 @@ func (m *Manager) Reconcile(request reconcile.Request) (reconcile.Result, error)
 	reqLogger := m.log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Certificates")
 
-	elapsedToRotate := m.elapsedToRotateFromLastDeadline()
+	elapsedToRotate := m.elapsedToRotateCAFromLastDeadline()
```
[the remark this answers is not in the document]
qinqon: Well we rotate CA and services in this case that's why I havent add any suffix, I can add "All".

pkg/certificate/controller.go
RamLavi: you're right, it's not our job to recreate them here. but the'll get the cert with the existing expiration. so no issue here, just wanted to make sure I understand

pkg/certificate/controller.go
RamLavi: I recline my comment :)

PR #37, opened 2020-07-23
Use a struct to group certificate options

qinqon:
The certificate options are starting to grow too much let's group them together
to pass the struct directly from client to certificate manager.

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -26,14 +26,13 @@ type Manager struct {
 	// client contains the controller-runtime client from the manager.
 	client client.Client
 
-	// webhookName The Mutating or Validating Webhook configuration name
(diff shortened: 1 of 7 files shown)


pkg/certificate/manager.go
```diff
@@ -86,23 +75,21 @@ const (
...
 	client client.Client,
-	webhookName string,
-	webhookType WebhookType,
-	namespace string,
-	caCertDuration time.Duration,
-	serviceCertDuration time.Duration,
+	options Options,
 ) *Manager {
 
+	options.setDefaults()
```
RamLavi, line 81: maybe also add a sanity check?

qinqon: I don't want to complicate this more than needed, we can add sanity at differet PR if needed.

PR #38, opened 2020-07-23
Implement CAOverlapInterval options to ditch old CAs from CABunlde

qinqon:
This commit add a knob to configure how much time time a CA certificate that is not
the latest generated on after rotation can live at
CABunlde instead of using expiration time, if it's not set it uses CARotateInterval
which is similar to expiration time. This reduce the size of CABundle.

### DIFF
--- a/pkg/certificate/cleanup.go
+++ b/pkg/certificate/cleanup.go
@@ -35,14 +35,18 @@ func (m *Manager) earliestCleanupDeadline() (time.Time, error) {
 }
 
 // earliestCleanupDeadlineForCACerts will inspect CA certificates
-// select the deadline based on certificate that is going to expire
-// sooner so cleanup is triggered then
+// select the deadline based on certificate NotBefore + caOverlapDuration
+// returning the daedline that is going to happend sooner
 func (m *Manager) earliestCleanupDeadlineForCerts(certificates []*x509.Certificate) time.Time {
-
 	var selectedCertificate *x509.Certificate
 
+	// There is no overlap just return expiration time
+	if len(certificates) == 1 {
+		return certificates[0].NotAfter
+	}
+
 	for _, certificate := range certificates {
-		if selectedCertificate == nil || certificate.NotAfter.Before(selectedCertificate.NotAfter) {
+		if selectedCertificate == nil || certificate.NotBefore.Before(selectedCertificate.NotBefore) {
 			selectedCertificate = certificate
 		}
 	}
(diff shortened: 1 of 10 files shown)


pkg/certificate/cleanup_test.go
```diff
@@ -283,7 +312,7 @@ var _ = Describe("CABundle cleanup", func() {
 				},
 				{
 					notBefore: -34 * time.Hour,
-					notAfter:  -33 * time.Hour,
+					notAfter:  90 * time.Hour,
 				},
```
RamLavi, line 316: what happens if all certs are past the cleaning time? should we remove all or keep at least one (if it's still valid of course)?

qinqon: Going to add a test for that, could happend that pod is down beyond the overlap duration and when it goes up it will clean al CABundle and that's too bad, it should keep the one with older issued date.

Makefile
```diff
@@ -2,6 +2,7 @@ SHELL := /bin/bash
 
 BIN_DIR = $(CURDIR)/build/_output/bin/
 
+export GOPROXY=off
```
RamLavi, line 5: is this relevant/used?

pkg/certificate/cleanup.go
```diff
@@ -72,14 +76,35 @@ func (m *Manager) cleanUpCABundle() error {
...
+		logger.Info("Checking certificate for cleanup", "now", now, "caOverlapDuration", m.caOverlapDuration, "NotBefore", certificate.NotBefore, "NotAfter", certificate.NotAfter)
+
+		// Expired certificate are cleaned up
+		caExpirationDate := certificate.NotAfter
+		if now.After(caExpirationDate) {
+			logger.Info("Cleaning up expired certificate", "now", now, "NotBefore", certificate.NotBefore, "NotAfter", certificate.NotAfter)
+			continue
+		}
+
+		// Certificates that are not the last appended one (the last generated from a rotation) and is pass CA Overlap Duration limit
```
RamLavi, line 100: please rephrase comment (suggestion below): 
`Clean up certificates that pass CA Overlap Duration limit, except for the last appended one (i.e. the last generated from a rotation) since we need at least one valid certificate`

pkg/certificate/cleanup_test.go
```diff
@@ -295,7 +324,52 @@ var _ = Describe("CABundle cleanup", func() {
 					notBefore: -1 * time.Hour,
 					notAfter:  80 * time.Hour,
 				},
+				{
+					notBefore: -34 * time.Hour,
+					notAfter:  90 * time.Hour,
+				},
 			},
 		}),
+		Entry("All beyond overlap duration, should keep the latest appended", cleanUpCertificatesCase{
```
RamLavi, line 333: please add another test where All are beyond overlap, and also expired.

Makefile
qinqon: Nope, removing it.

PR #42, opened 2020-10-14
Bump to go 1.15 to get SubjectKeyId at CA certificate

qinqon:
Go crypto library for versions < 1.15 is not filling in the SKI field for CAs [1] so we bump to 1.15
[1] golang/go#26676
[2] https://go-review.googlesource.com/c/go/+/227098/

Closes: #42

### DIFF
--- a/pkg/certificate/triple/triple_test.go
+++ b/pkg/certificate/triple/triple_test.go
@@ -1,0 +1,45 @@
+package triple
+
+import (
+	"crypto/x509"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Cert library", func() {
+	Context("when NewCA is called", func() {
+		var (
+			name     string
+			duration time.Duration
+			now      time.Time
+		)
+		BeforeEach(func() {
+			now = time.Now()
+			name = "foo-bar-name"
+			duration = time.Minute
+			Now = func() time.Time { return now }
+		})
+		It("should generate key and CA cert with expected fields", func() {
+
+			keyAndCert, err := NewCA(name, duration)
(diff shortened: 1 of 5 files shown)


pkg/certificate/triple/cert.go
```diff
@@ -80,6 +90,13 @@ func NewSelfSignedCACert(cfg Config, key crypto.Signer, duration time.Duration)
 		BasicConstraintsValid: true,
 		IsCA:                  true,
 	}
+
+	// Golang < 1.15 is not filling in the SKI field for CAs [1] so we replicate the golang >= 1.15 fix [2]
```
rhrazdil, line 94: Are we okay with overwriting `tml.SubjectKeyId` when we move to golang >= 1.15? Shouldn't we check `tmpl.SubjectKeyId` is empty first?

qinqon: That was my original idea but it means that it will call CreateCertificate twice at golang < 1.15 another option is to check the golang version but that's even worse and the third option is to do nothing and just try to bump to 1.15 all the parts.

rhrazdil: ahh, yeah gotcha.

RamLavi (commented):
small question

pkg/certificate/triple/triple_test.go
```diff
@@ -0,0 +1,45 @@
...
+		var (
+			name     string
+			duration time.Duration
+			now      time.Time
+		)
+		BeforeEach(func() {
+			now = time.Now()
+			name = "foo-bar-name"
+			duration = time.Minute
+			Now = func() time.Time { return now }
```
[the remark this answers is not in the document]
qinqon: look at `now = time.Now()` this is initialize only once so we return allways the same time and we can check.

RamLavi (approved):
/lgtm

PR #43, opened 2021-02-10
Parameterize overlap at certificate cleanup

qinqon:
The current cleanup logic is stick to CABundle to be able to calculate
cleanup for service certs we can parameterize this logic so we pass
CA or services certificates with its corresponding overlaping values.

Also now the "overlap" option means the time before "NotAfter" to start rotating the certificates, the certificates cleanup is down at certificate expiration.

Test PRs:
kmp: https://github.com/k8snetworkplumbingwg/kubemacpool/pull/272
knmstate: https://github.com/nmstate/kubernetes-nmstate/pull/696


pkg/certificate/cleanup.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
```
RamLavi, line 6: so, in the `Refactor overlap to use it ony at rotate` commit (btw typo in ony-->only)
we change from doing cleanp when notBefore+overlap to cleanup when noAfter?
and the overlap will only be used now to discover the rotation itself?

pkg/certificate/cleanup_test.go
```diff
@@ -243,16 +233,15 @@ var _ = Describe("CABundle cleanup", func() {
...
+		Entry("middle one expired right now, should remove it", cleanUpCertificatesCase{
 			certsExpiration: []certificateExpiration{
 				{
 					notBefore: -1 * time.Hour,
 					notAfter:  33 * time.Hour,
 				},
 				{
 					notBefore: -5 * time.Hour,
-					notAfter:  44 * time.Hour,
+					notAfter:  -1 * time.Hour,
```
RamLavi, line 244: shouldn't `expired right now` mean that the notAfter should be 0?

pkg/certificate/cleanup_test.go
```diff
@@ -322,37 +309,36 @@ var _ = Describe("CABundle cleanup", func() {
...
 				},
-				{
-					notBefore: -34 * time.Hour,
-					notAfter:  90 * time.Hour,
-				},
 			},
 		}),
-		Entry("All beyond overlap duration, should keep the latest appended", cleanUpCertificatesCase{
-			overlapDuration: 5 * time.Hour,
+		Entry("One certificate expired now, should remove it", cleanUpCertificatesCase{
```
RamLavi, line 314: I think you have duplicate Entry here.

pkg/certificate/secret.go
```diff
@@ -51,11 +51,25 @@ func addTLSCertificate(data map[string][]byte, cert *x509.Certificate) error {
...
+func setAnnotation(secret *corev1.Secret) {
+	if secret.Annotations == nil {
+		secret.Annotations = map[string]string{}
+	}
+	secret.Annotations[secretManagedAnnotatoinKey] = ""
 }
 
-func populateTLSSecret(secret corev1.Secret, keyPair *triple.KeyPair) (*corev1.Secret, error) {
+func resetTLSSecret(secret corev1.Secret, keyPair *triple.KeyPair) (*corev1.Secret, error) {
+
```
RamLavi, line 62: extra line

pkg/certificate/controller_test.go
```diff
@@ -126,15 +127,21 @@ var _ = Describe("Certificates controller", func() {
 		var (
 			previousResult, currentResult reconcile.Result
 			previousTLS, currentTLS       TLS
+			backToTheFuture               = func(step string, future time.Duration) {
```
RamLavi, line 130: iiuc, this function has an isolation problem, because if you run it twice the "now" changes, this can be confusing since the "now" is not fixed...
can you set the "now" relative to a fixed "now" that is declared outside of this function?

RamLavi (changes requested):
first pass, added comments and questions..

pkg/certificate/cleanup.go
```diff
@@ -65,7 +59,7 @@ func (m *Manager) cleanUpCABundle() error {
 		if err != nil {
 			return nil, errors.Wrap(err, "failed getting ca certs to start cleanup")
 		}
-		cleanedCAs := m.cleanUpCertificates(cas)
+		cleanedCAs := m.cleanUpCertificates(cas, m.caOverlapDuration)
```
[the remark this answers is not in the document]
qinqon: This is old change, at the end we will just do the cert cleanup at expiration time of the old certificate so they will be overlapping `oldcert.NotAfter - newcert.NotBefore`

pkg/certificate/cleanup.go
```diff
@@ -99,9 +93,9 @@ func (m *Manager) cleanUpCertificates(certificates []*x509.Certificate) []*x509.
 
 		// Clean up certificates that pass CA Overlap Duration limit,
```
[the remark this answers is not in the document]
qinqon: This is old we trigger cleanup at certificate expiration.

pkg/certificate/controller_test.go
```diff
@@ -186,22 +192,27 @@ var _ = Describe("Certificates controller", func() {
...
-						By("Reconcile at second service cert rotation deadline")
+						By("Reconcile at service certificate overlap cleanup")
 						currentResult, err = mgr.Reconcile(reconcile.Request{})
 						Expect(err).To(Succeed(), "should success reconciling")
 						currentTLS = getTLS()
 					})
 					It("should rotate service cert/key and return a new deadline", func() {
 						Expect(currentTLS.serviceCertificate).ToNot(Equal(previousTLS.serviceCertificate), "should have do TLS cert rotation")
-						Expect(currentTLS.servicePrivateKey).ToNot(Equal(previousTLS.servicePrivateKey), "should have do a TLS key rotation")
+						Expect(currentTLS.servicePrivateKey).To(Equal(previousTLS.servicePrivateKey), "should have do a TLS key rotation")
```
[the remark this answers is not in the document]
qinqon: `It` message is wrong this cycle is about service cert cleanup, the private key is keep, I will change it, thanks!.

pkg/certificate/secret.go
```diff
@@ -184,7 +221,31 @@ func (m *Manager) getTLSKeyPair(secretKey types.NamespacedName) (*triple.KeyPair
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed parsing TLS private key PEM at secret %s", secretKey)
 	}
-	return &triple.KeyPair{Key: privateKey.(*rsa.PrivateKey), Cert: certs[0]}, nil
+
+	// Certs are appended to implement overlap so we take the last one
+	// it will match with the key
+	lastAppendedCert := certs[len(certs)-1]
```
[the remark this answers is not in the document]
qinqon: Totally

pkg/certificate/cleanup.go
qinqon: Same idea as `renewBefore` here https://docs.cert-manager.io/en/release-0.8/reference/certificates.html.

pkg/certificate/cleanup_test.go
qinqon: Yep, fixing!!!

pkg/certificate/cleanup_test.go
qinqon: I don't see it duplicated.

pkg/certificate/cleanup.go
```diff
@@ -10,34 +10,28 @@ import (
 	"github.com/qinqon/kube-admission-webhook/pkg/certificate/triple"
 )
```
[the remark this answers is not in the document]
qinqon: Yep

pkg/certificate/secret.go
qinqon: Done

pkg/certificate/controller_test.go
qinqon: This is the whole purpose, calling it multiple time throw you to the future.

pkg/certificate/cleanup.go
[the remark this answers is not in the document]
RamLavi: ack.
Like on the time diagram :)

pkg/certificate/cleanup.go
RamLavi: then why not call it `renewBefore`

pkg/certificate/cleanup_test.go
RamLavi: I can't find it either now.. maybe my mistake sorry

pkg/certificate/controller_test.go
RamLavi: let me explain with example: 
- first call `backToTheFuture("Reconcile for the first time", 0)`: mgr.now() and my watch have 0 difference
- second call `backToTheFuture("Reconcile in the middle of service cert deadline", serviceCertDuration/2)`: mgr.now() adds `duration/2` so it is after my watch, so far so good.
- third call `backToTheFuture("Reconcile at service cert rotation deadline", currentResult.RequeueAfter)`: mgr.now() adds `duration/2+currentResult.RequeueAfter`, and not `currentResult.RequeueAfter`...

in other words,
```
now = mgr.now().Add(future)
mgr.now = func() time.Time { return now }
```
so now always adds to the last mgr.now that was set earlier... this is what I mean by isolation issue.

pkg/certificate/cleanup.go
qinqon: It will break the API but I supose is not important and it's better to understand it, I will put a commit in place to rename it.

pkg/certificate/controller_test.go
qinqon: I still don't get it, this is a time machine and the offset is relative to the time you have moved to.

RamLavi (changes requested):

pkg/certificate/controller_test.go
```diff
@@ -175,161 +172,184 @@ var _ = Describe("Certificates controller", func() {
...
-							Expect(currentTLS.caPrivateKey).ToNot(Equal(previousTLS.caPrivateKey), "should have rotate CA key rotation")
-
-							elapsedForCleanup, err := mgr.earliestElapsedForCleanup()
-							Expect(err).To(Succeed(), "should succeed calculating earliestElapsedForCleanup")
-							Expect(currentResult.RequeueAfter).To(Equal(elapsedForCleanup), "Reconcile at rotate should schedule next Reconcile to do the CA overlapping cleanup")
-
-							cas, err := triple.ParseCertsPEM(currentTLS.caBundle)
-							Expect(err).To(Succeed(), "should succeed parssing caBundle")
-							Expect(cas).To(HaveLen(2), "should overlap CAs")
+							Expect(currentTLS.caBundle).To(Equal(previousTLS.caBundle), "shouldn't have rotate CABundle ")
```
RamLavi, line 210: These multiple expects are repeating themselves, what do you think about moving them to 4 functions:
- `CheckServiceRotated(previousTLS, currentTLS)`
- `CheckServiceNotRotated(previousTLS, currentTLS)`
- `CheckCARotated(previousTLS, currentTLS)`
- `CheckCANotRotated(previousTLS, currentTLS)`

RamLavi (changes requested):

pkg/certificate/cleanup.go
RamLavi: no problem to do it in a separate PR

qinqon: ack, thanks.

RamLavi (approved):

PR #46, opened 2021-02-23
Use webhookconfig v1

qinqon:
The admissionwebhook v1beta1 is going to be deprecated at k8s 1.22 [1]. This change bumps controller-runtime to 0.8.2 to support k8s >= 1.20 and bump admissionwebhook configuration to v1.

[1] https://github.com/kubernetes/kubernetes/pull/79549


pkg/webhook/server/server.go
```diff
@@ -137,15 +138,15 @@ func (s *Server) waitForTLSReadiness() error {
...
+func (s *Server) Start(ctx context.Context) error {
 	s.log.Info("Starting nodenetworkconfigurationpolicy webhook server")
 
 	err := s.waitForTLSReadiness()
 	if err != nil {
 		return errors.Wrap(err, "failed watting for ready TLS key/cert")
 	}
 
-	err = s.webhookServer.Start(stop)
+	err = s.webhookServer.Start(ctx)
```
RamLavi, line 149: what happened to the stop channel?

RamLavi, line 149: so you can now add a context cancel function instead of the stop channel? (just asking, not suggesting you need to add such a function)

qinqon: You must, they hare replace one for another, after context concept appeart at golang is the most idiomatic way of do the stop ching.

RamLavi:
hey @qinqon why did you remove the test/pod folder files?
isn't it used as an example for whoever will want to use this library?

qinqon:
> hey @qinqon why did you remove the test/pod folder files?
> isn't it used as an example for whoever will want to use this library?

If you want to see how to use it just check the server tests I would say.

RamLavi (approved):
/lgtm

PR #47, opened 2021-04-07
Support multiple webhooks & refactor

jcaamano:
The primary concern of this commit is to add support for handling
multiple webhooks with a single certificate manager/controller.
The final idea is to set this as an independent library apt for external use.

This lead to a refactor to isolate certificate handling from
configuration handling.

A new package certificate/chain defines data structures to represent a
certificate chain and to ensure that certificate chain is current in
regards to expiration and overlap times, rotating where necessary.

All configuration handling has been refactored in configuration.go with
a generic concept of mapping runtime objects to/from certificate chain.

Other functional changes include:

* Drop support for URL client config in webhooks. It does not make sense
to manage certificates for webhooks that are hosted outside of the
cluster.
* Removed retries. Assumed that the primary retry mechanism should be to
requeue a controller request versus handling retries internally.


jcaamano:
Since this a big refactor, I suggest doing a staggered review. Let's focus first on APIs and usage:

* `pkg/certificate/chain/api.go` and usage in `pkg/certificate/controller.go` @ `Reconcile`
* `pkg/certificate/types.go`, `pkg/certificate/manager.go` and usage in `pkg/webhook/server.go`@ New & WithHook.

qinqon:
> Since this a big refactor, I suggest doing a staggered review. Let's focus first on APIs and usage:
> 
> * `pkg/certificate/chain/api.go` and usage in `pkg/certificate/controller.go` @ `Reconcile`
> * `pkg/certificate/types.go`, `pkg/certificate/manager.go` and usage in `pkg/webhook/server.go`@ `New` & `WithHook`

How do you feel about splitting with commits, so it's easier to review.

jcaamano:
> How do you feel about splitting with commits, so it's easier to review.

I agree that this is a one big commit that would be better if split in smaller ones. But as I started to work on this with refactoring in mind it would have required more work in terms of temporary,intermediate code. I think at this time is more practical to do a staggered guided review if possible.

pkg/certificate/controller.go
```diff
@@ -87,107 +102,31 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-		// Re-calculate elapsedToRotateServices since we have generated new
-		// services certificates
-		m.nextRotationDeadlineForServices()
-		elapsedToRotateServices = m.elapsedToRotateServicesFromLastDeadline()
-	}
+	objects := objectMap{}
+	certificateChain := chain.CertificateChain{}
 
-	elapsedForCABundleCleanup, err := m.earliestElapsedForCACertsCleanup()
+	err := m.readCertificateChain(objects, &certificateChain)
```
qinqon, line 110: What do you think about moving al the Code from  `Reconcile` to `Manager.reconcileCertificates` this way we cleanup a little the already bloated controller (a lot of this code is boilerplate)

Apart from that why don't we make `readCertificateChain` return directly objects and chain instead of needed them as arguments

pkg/certificate/types.go
```diff
@@ -0,0 +1,21 @@
...
+	ValidatingWebhook WebhookType = "Validating"
+)
+
+type WebhookReference struct {
+	Type WebhookType
+	Name string
+}
+
+func (w WebhookReference) String() string {
+	return fmt.Sprintf("%s/%s", w.Type, w.Name)
```
qinqon, line 20: Let's add "Webhook" suffix in case we print this out of webhook context 

```suggestion
	return fmt.Sprintf("%sWebhook/%s", w.Type, w.Name)
```

pkg/webhook/server/server.go
```diff
@@ -21,65 +20,75 @@ import (
...
+	"github.com/qinqon/kube-admission-webhook/pkg/certificate/chain"
 )
 
 type Server struct {
-	webhookServer *webhook.Server
-	certManager   *certificate.Manager
-	log           logr.Logger
+	webhookServer  *webhook.Server
+	webhookConfigs map[certificate.WebhookReference]bool
+	certManager    func() (*certificate.Manager, error)
```
qinqon, line 29: To follow controller-runtime way https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/manager.go#L227 let's call this `newCertManager` since it's a builder function.

pkg/webhook/server/server.go
```diff
@@ -21,65 +20,75 @@ import (
...
+		return certManager, err
+	}
+
 	return s, nil
 }
 
-func WithHook(path string, hook *webhook.Admission) ServerModifier {
-	return func(s *webhook.Server) {
-		s.Register(path, hook)
+func WithHook(path string, hook *webhook.Admission, config certificate.WebhookReference) ServerModifier {
```
qinqon, line 60: There is no real relation between hooks and config, in case we have the same hook and is used a two different config dos it means that we have to call `WithHook` twice, so let's add a `WithConfig`

pkg/webhook/server/server.go
```diff
@@ -104,26 +113,6 @@ func (s *Server) checkTLS() error {
...
-	caPEM, err := s.certManager.CABundle()
-	if err != nil {
-		return errors.Wrap(err, "failed to retrieve CA cert")
-	}
-
-	err = triple.VerifyTLS(certPEM, key, caPEM)
-	if err != nil {
-		return errors.Wrapf(err, "failed verifying %s/%s", certsPath, keyPath)
-	}
-
```
qinqon, line 126: we have to verify the chain here

qinqon (changes requested):
First pass, reviewed what is specified at https://github.com/qinqon/kube-admission-webhook/pull/47#issuecomment-814814617

pkg/certificate/chain/api.go
```diff
@@ -0,0 +1,80 @@
...
+	// of them will share the same duration
+	CertRotateInterval time.Duration
+
+	// CertOverlapInterval the duration of service certificates at bundle if
+	// not set it will default to CertRotateInterval
+	CertOverlapInterval time.Duration
+}
+
+// Update keeps the certificate chain data currrent by:
+// - Rotating all issued certificates when at least one is expired or in the
```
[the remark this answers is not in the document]
jcaamano: I am not sure I get this reasoning. Certificates would share the same expiration time and overlap time but be rotated at different times if they were issued at different times, which is something that might happen supporting multiple webhooks when not all of the exist initially.

The only reasoning to rotate them at the same time is convenience of implementation as far as I can see.

pkg/certificate/controller.go
jcaamano: > What do you think about moving al the Code from `Reconcile` to `Manager.reconcileCertificates` this way we cleanup a little the already bloated controller (a lot of this code is boilerplate)

OK

> Apart from that why don't we make `readCertificateChain` return directly objects and chain instead of needed them as arguments

It is more clear to me the usage of this kind of interface that the intention of these methods (`read/update/write`) is to make modifications in place, even though for some might not apply. The commonality of it all makes it clear that everything happens in place of those to arguments throughout all operations.

pkg/webhook/server/server.go
jcaamano: Why?

If anything, it seems arbitrary to do it only just once at start knowing that the certificates might be rotated later on. I don't even know why we need to check for the key/cert files. The underlying server implementation already has a fsnotify watcher doing it.

qinqon: If we start the webhook server with the "empty" placeholder secret to allow pod to start the pod will fail and go to restart mode and that's problematic.

we fixed it by checking TLS before starting the webhook server.

Remember that cert rotation mechanism runs in the same pod as webhook server so we have a chicken egg issue here and we create a placeholder secret with garbage TLS key/cert so pod can start.

jcaamano: OK, I am not absolutely convinced that the server should fail & restart missing the TLS key/cert, that is probably worth a revisit.

Anyway, we could call `certManager.VerifyTLS` instead. We probably need to mutex that against `certManager.reconcileCertificates` to be on the safe side. Thoughts?

qinqon: Since we are accessing already the keys instead of using certManager.VerifyTLS let's create a new chain reading the files and call verifyTLS on the chain, so we don't need a mutex.

jcaamano: A new chain is already being used, that's not the reason of the mutex. The reason would be non transactional DB/cache.

jcaamano: `verifyTLS` is read only, the mutex would just avoid another poll cycle.

qinqon: What do you mean ?

jcaamano: `checkTLS` is called with a polling function every 5 secs for 5 min.

jcaamano: So additionally to what we talked about @ https://github.com/qinqon/kube-admission-webhook/pull/47#discussion_r610488157, I went for the compromise of adding a flag to manager when verifying so that if active then all provided webhooks must be read successfully and thus cascade on a complete successful verification.

So reconcile is convergent in that not all webhook configs need to exist at once but reconciliation happens for those that do without errors, while verification requires that a valid certificate chain exists for all webhook configs. 

Check `manager.go`& `server.go`

jcaamano:
Ok, moving on we can review `certificate/chain` package. It contains all the certificate specific functionality and it should be functionally equivalent to the previous implementation and also have a very similar code structure, highlighting the following differences:
* `api.go` contains (new) exported types & functions
* `chain.go` deals with handling those new types which are extended with unexported variants with the decoded key/certs (this allows for cleaner error handling), setters/getters, etc...
* `deadlines.go`. `cleanup.go`, `rotation.go` is equivalent to previous functionality, just stuff moved around.
* All expiration calculations are based on `Time`, `Duration` is not used as that is only a concern for the controller. So this package returns the earliest time next update should happen.
* No expiration times are cached as before. They are recalculated on every update. I thought that the optimization wasn't worth it vs code complexity.

pkg/certificate/chain/chain.go
```diff
@@ -0,0 +1,288 @@
...
+	for i := range data.CertificatesIssued {
+		certificateIssue := certificateIssue{
+			CertificateIssue: data.CertificatesIssued[i],
+			caCerts:          make(map[string][]*x509.Certificate),
+		}
+
+		key, certs, err := keyPairPemToKeypair(certificateIssue.KeyPEM, certificateIssue.CertPEM)
+		certificateIssue.key = key
+		certificateIssue.certs = certs
+		if err != nil {
```
qinqon, line 76: Catch the error before assigning stuff.

pkg/certificate/chain/chain.go
```diff
@@ -0,0 +1,288 @@
...
+			logger.Info("Certificate key pair invalid, will force all issued certificates rotation",
+				"name", certificateIssue.Name,
+				"err", err,
+			)
+		}
+
+		for k, v := range certificateIssue.CACertPEM {
+			certs, err := triple.ParseCertsPEM(v)
+			certificateIssue.caCerts[k] = certs
+			if err != nil {
```
qinqon, line 88: ditto

pkg/certificate/chain/chain.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func (r *certificateChain) update() (time.Time, error) {
+	logger := r.log.WithName("update")
+	logger.Info("Checking certificate chain for rotation or cleanup")
+
+	deadlineToRotateCA := r.findRotationDeadlineForCA()
+	deadlineToRotateCerts := r.findRotationDeadlineForCerts()
+	rotateCA := !r.now().Before(deadlineToRotateCA)
+	rotateCerts := !r.now().Before(deadlineToRotateCerts)
```
qinqon, line 195: Is more clear if we use "After" than negate "Before"

pkg/certificate/chain/chain.go
```diff
@@ -0,0 +1,288 @@
...
+
+	// Ensure certificate chain
+	if !rotateCA {
+		err := r.verifyTLS()
+		if err != nil {
+			logger.Info("Certificate chain failed verification, will force full chain rotation", "err", err)
+			// Force rotation
+			rotateCA = true
+		}
+	}
```
qinqon, line 205: If a service certificate expire it will enter here and all the certificates will be rotated, we don't need this block.

qinqon (changes requested):
First pass to cleanup structures

pkg/certificate/chain/api.go
```diff
@@ -0,0 +1,78 @@
...
+	CertPEM   []byte
+	CACertPEM map[string][]byte
+}
+
+// CA contains details about a certification authority
+type CA struct {
+	Name    string
+	KeyPEM  []byte
+	CertPEM []byte
+}
```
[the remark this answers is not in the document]
jcaamano: Done!

pkg/certificate/chain/chain.go
```diff
@@ -0,0 +1,288 @@
...
+	*CertificateIssue
+	key     *rsa.PrivateKey
+	certs   []*x509.Certificate
+	caCerts map[string][]*x509.Certificate
+}
+
+type ca struct {
+	*CA
+	keyPair *triple.KeyPair
+}
```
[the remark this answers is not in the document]
jcaamano: Done!

pkg/certificate/chain/chain.go
jcaamano: It's not equivalent.

pkg/certificate/chain/chain.go
jcaamano: Just as it was being done before, I am not changing functionality here.

pkg/certificate/chain/chain.go
jcaamano: The error is just logging.

pkg/certificate/chain/chain.go
jcaamano: The error is just logging.

pkg/certificate/chain/chain.go
qinqon: Before don't return true on equals so !Before is equivalent to After and expressing this as "if now is after deadline" is more clear.

pkg/certificate/chain/chain.go
qinqon: Right... and now I see it was wrong, let's fix it at follow up PR.

pkg/certificate/chain/api.go
```diff
@@ -0,0 +1,90 @@
...
+
+// CertificateIssue contains details about an issued certificate, including the
+// private key and named CA certificates known to verify the issued certificate.
+type CertificateIssue struct {
+	Name      string
+	IPs       []string
+	Hostnames []string
+	KeyPEM    []byte
+	CertPEM   []byte
+	CACertPEM map[string][]byte
```
qinqon, line 19: I have to go back to this struct, why do we need multiple CACerts per caBundle ? all the CABundle will share the same content I think we don't need a map here.

jcaamano: Ensuring that all the ca bundles share the some content requires specific logic that was not there in the previous implementation either (where it would only append, not replace, to the ca bundle). That logic would be now more relevant & complicated given that we support multiple webhook configs with possibly different lifetimes. So in this case I think it is more useful & simple to have a faithful representation of the data structure, which allows for further improvements in the future, let's say for example if we would like to handle different webhooks at different times as well.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+// readCertificateChain is the entry point to read from K8s all the certificate
+// chain data related to the managed webhooks. This information is both tracked
+// as a K8s object map and a certificate chain specific structure which should
+// be supplied to this method initialized and empty.
+func (m *Manager) readCertificateChain(objects objectMap, certificateChain *chain.CertificateChainData) error {
+	m.initObjects(objects)
+	certificateChain.CA.Name = m.secretCAName().String()
+	err := m.readObjectsToChain(objects, certificateChain)
+	return err
 }
```
qinqon, line 97: What do you think about changing this into
```suggestion
func (m *Manager) readCertificateChain(objects objectMap) (*chain.CertificateChain, error) {
	m.initObjects(objects)
	certificateChain := newChain()
	certificateChain.CA.Name = m.secretCAName().String()
	err := m.readObjectsToChain(objects, certificateChain)
	if err != nil {
	    return nil, err
	}
	return certificateChain, nil
}
```

Then we don't need to pass Options to `CertificateChain` we just have a Update Receiver on it that pass the Update options, this way we can remove `certificateChain` (removing also de logs and the timer since it can be set globally at triple.time)

qinqon (changes requested):
We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`

pkg/certificate/chain/chain.go
jcaamano: Right so
!Before is equivalent to equal or after
After is just after
!Before is not equivalent to After

pkg/certificate/chain/api.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// decoded data
+	key     *rsa.PrivateKey
+	certs   []*x509.Certificate
+	caCerts map[string][]*x509.Certificate
```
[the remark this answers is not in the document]
jcaamano: Same as before: `type NamedCACerts map[string][]*x509.Certificate`and `caCerts NamedCACerts` ?

pkg/certificate/chain/chain.go
qinqon: Ahh! you right, all good then, thanks for clarification.

pkg/certificate/chain/api.go
```diff
@@ -0,0 +1,90 @@
...
+	CertPEM []byte
+
+	// decoded data
+	keyPair *triple.KeyPair
+}
+
+// CertificateChainData represents details about a certification authority and
+// named certificates issued by that authority.
+type CertificateChainData struct {
+	CertificatesIssued map[string]*CertificateIssue
```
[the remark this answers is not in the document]
jcaamano: Again, did not want to introduce the specific Service meaning to a generic certificate chain concept. These are just issued certificates. Can we come up with a more clear & generic name perhaps?

pkg/certificate/chain/api.go
qinqon: We don't want generic mapping this is an opinionated cert-rotation lib desgined for addmission webhooks so referencing CABundleName is ok.

Also maps naming maps with ValueByKey makes clear is a map while KeydValue could be a slice.

pkg/certificate/chain/api.go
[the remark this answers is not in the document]
qinqon: ditto https://github.com/qinqon/kube-admission-webhook/pull/47#discussion_r612307621

pkg/certificate/chain/api.go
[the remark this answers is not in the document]
qinqon: ditto https://github.com/qinqon/kube-admission-webhook/pull/47#discussion_r612307621

pkg/certificate/chain/api.go
jcaamano: I understand that but that completely breaks a design decision I made when refactoring that was abstracting the certificate chain rotation. So to be sensible to this approach, is there no alternative possibility to clarify, either by more clear generic naming or more doc what this is, as a generalized concept? Because otherwise we are up for renaming everything from now on, reversing all considerations that I had while working on this.

qinqon: Ack, let's keep this generic then, but let's use "ByName" since it's a known golang pattern to name maps this way.

jcaamano: great!

jcaamano:
> We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`

We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?

jcaamano:
> > We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`
> 
> We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?

Also certificateChain has additional things vs CertificateChainData. Not sure if this comment was old though and no longer applies.

qinqon:
> > We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`
> 
> We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?

So we can have a constructor that you pass CA and CertificatesIssued so 

```golang
certificateChain, err := newEncodedChain(ca, certificateIssued) //it will compose the struct and encode the chain
```

qinqon:
> > > We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`
> > 
> > 
> > We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?
> 
> Also certificateChain has additional things vs CertificateChainData. Not sure if this comment was old though and no longer applies.

the extra things are not really needed:
- log (we can live a global log)
- time mock (we can just use triple.time and reset it) 
- options we don't need them since they are passed to Update/Validate

jcaamano:
> > > > We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`
> > > 
> > > 
> > > We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?
> > 
> > 
> > Also certificateChain has additional things vs CertificateChainData. Not sure if this comment was old though and no longer applies.
> 
> the extra things are not really needed:
> 
> * log (we can live with global log)
> * time mock (we can just use triple.time and reset it on tests)
> * options we don't need them since they are passed to Update/Validate
> 
> So the idea is to remove current `certificateChain` and to rename `CertificateChainData` to `certificateChain` that can be constructed by `newEncodedChain` then add `Update` and `Validate` receivers on it.

Chain data has to be read before being updated/validated. Different types provide this separation of concern seamlessly. Same type for both things is not a clear interface from my point of view and requires specific code to prevent misuse. Am I missing something? Are we trying to oversimplify things?

qinqon:
> > > > > We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`
> > > > 
> > > > 
> > > > We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?
> > > 
> > > 
> > > Also certificateChain has additional things vs CertificateChainData. Not sure if this comment was old though and no longer applies.
> > 
> > 
> > the extra things are not really needed:
> > 
> > * log (we can live with global log)
> > * time mock (we can just use triple.time and reset it on tests)
> > * options we don't need them since they are passed to Update/Validate
> > 
> > So the idea is to remove current `certificateChain` and to rename `CertificateChainData` to `certificateChain` that can be constructed by `newEncodedChain` then add `Update` and `Validate` receivers on it.
> 
> Chain data has to be read before being updated/validated. Different types provide this separation of concern seamlessly. Same type for both things is not a clear interface from my point of view and requires specific code to prevent misuse. Am I missing something? Are we trying to oversimplify things?

We are already aggregating PEM and Struct cert versions on same structures embedding before the review and directly after the review also the only difference is PEM vs Certificate struct, in fact we can even use only the Certificate struct at the chain and do the coding/encoding at functions when needed.

jcaamano:
> > > > > > We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`
> > > > > 
> > > > > 
> > > > > We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?
> > > > 
> > > > 
> > > > Also certificateChain has additional things vs CertificateChainData. Not sure if this comment was old though and no longer applies.
> > > 
> > > 
> > > the extra things are not really needed:
> > > 
> > > * log (we can live with global log)
> > > * time mock (we can just use triple.time and reset it on tests)
> > > * options we don't need them since they are passed to Update/Validate
> > > 
> > > So the idea is to remove current `certificateChain` and to rename `CertificateChainData` to `certificateChain` that can be constructed by `newEncodedChain` then add `Update` and `Validate` receivers on it.
> > 
> > 
> > Chain data has to be read before being updated/validated. Different types provide this separation of concern seamlessly. Same type for both things is not a clear interface from my point of view and requires specific code to prevent misuse. Am I missing something? Are we trying to oversimplify things?
> 
> We are already aggregating PEM and Struct cert versions on same structures embedding before the review and directly after the review also the only difference is PEM vs Certificate struct, in fact we can even use only the Certificate struct at the chain and do the coding/encoding at functions when needed.

I am not sure I understand, I need more detail. There is many different ways this can be done. I am trying to reason with mine. I already explained why I did the coding/encoding early and thought the explanation was valid and we agreed upon it. If we want to do things differently we need to find value because it is going to require extra work. I understand you probably would have done things differently.

We need to go back and review `pkg/certificate/chain/api.go` again and stick to it.

What is already proposed in this file is a clear separation of concern on where a user should load & provide the data `CertificateChainData` and how this data is then validated and decoded in a way that does not concern the user in `newChain`. The aggregation of fields we did don't matter as they are not exported, I agreed to that.

On to your proposal, unexported `certificateChain` and `newEncodedChain` but exported `Update` and `Validate` receivers? 
How does the user obtain a `certificateChain` to read data into? Why exported `Update` and `Validate` receivers? When is decoding and validation happening?

qinqon:
> > > > > > > We can simplify a little more and remove `certificateChain` And just have Update and Validate as Receivers of `CertificateChain`
> > > > > > 
> > > > > > 
> > > > > > We miss the possibility to have a constructor then. What should `newChain` receive and what should it return?
> > > > > 
> > > > > 
> > > > > Also certificateChain has additional things vs CertificateChainData. Not sure if this comment was old though and no longer applies.
> > > > 
> > > > 
> > > > the extra things are not really needed:
> > > > 
> > > > * log (we can live with global log)
> > > > * time mock (we can just use triple.time and reset it on tests)
> > > > * options we don't need them since they are passed to Update/Validate
> > > > 
> > > > So the idea is to remove current `certificateChain` and to rename `CertificateChainData` to `certificateChain` that can be constructed by `newEncodedChain` then add `Update` and `Validate` receivers on it.
> > > 
> > > 
> > > Chain data has to be read before being updated/validated. Different types provide this separation of concern seamlessly. Same type for both things is not a clear interface from my point of view and requires specific code to prevent misuse. Am I missing something? Are we trying to oversimplify things?
> > 
> > 
> > We are already aggregating PEM and Struct cert versions on same structures embedding before the review and directly after the review also the only difference is PEM vs Certificate struct, in fact we can even use only the Certificate struct at the chain and do the coding/encoding at functions when needed.
> 
> I am not sure I understand, I need more detail. There is many different ways this can be done. I am trying to reason with mine. I already explained why I did the coding/encoding early and thought the explanation was valid and we agreed upon it. If we want to do things differently we need to find value because it is going to require extra work. I understand you probably would have done things differently.
> 
> We need to go back and review `pkg/certificate/chain/api.go` again and stick to it.
> 
> What is already proposed in this file is a clear separation of concern on where a user should load & provide the data `CertificateChainData` and how this data is then validated and decoded in a way that does not concern the user in `newChain`. The aggregation of fields we did don't matter as they are not exported, I agreed to that.
> 
> On to your proposal, unexported `certificateChain` and `newEncodedChain` but exported `Update` and `Validate` receivers?
> How does the user obtain a `certificateChain` to read data into? Why exported `Update` and `Validate` receivers? When is decoding and validation happening?

After some thinking It's ok how it is since Verify and Update are going to be call at different times, so let's continue with need review stage.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
-	} else if m.webhookType == ValidatingWebhook {
-		validatingWebhookConfig := validatingWebhookConfig(webhook)
-		for i, _ := range validatingWebhookConfig.Webhooks {
-			clientConfig := &validatingWebhookConfig.Webhooks[i].ClientConfig
-			clientConfigList = append(clientConfigList, clientConfig)
-		}
+// keyedObject references a K8s resource object by key
+type keyedObject struct {
+	key     *objectKey
+	kobject client.Object
```
qinqon, line 56: Can we name this "obj" is what is used around

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+}
+
+// mapWebhookToChain maps a webhook object to certificate chain data.
+// If the webhook has no relevant data or does not exist, it is removed
+// from the object map so that is no longer considered. For every backing
+// service of the webhook, a reference to the service secret is added to the
+// object map if not already there.
+func mapWebhookToChain(object *keyedObject, objects objectMap, certificateChain *chain.CertificateChainData) {
+	clientConfigMap := clientConfigMap(object.kobject)
+	if len(clientConfigMap) <= 0 {
```
qinqon, line 255: len cannot be negative.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
-		for _, clientConfig := range m.clientConfigList(webhook) {
-			// Update the CA bundle at webhook
-			updatedCABundle, err := updateCABundle(clientConfig.CABundle)
-			if err != nil {
-				return errors.Wrap(err, "failed updating CA bundle")
-			}
-			clientConfig.CABundle = updatedCABundle
+		caBundleName := caBundleName(object.key.String(), name)
+		certificateChain.CertificatesIssued[serviceHostname].CACertPEM[caBundleName] = config.CABundle
+		key := newObjectKey("Secret", serviceNamespace, serviceName)
```
qinqon, line 275: We can use "secretType" constant from https://github.com/qinqon/kube-admission-webhook/pull/47/files#diff-3786f9aede730888d9ae80e4f57f664d73fa419e35ea86cdfffea2335cee65c9R40

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+		if _, found := objects[key]; !found {
+			objects[key] = &keyedObject{key, nil}
```
qinqon, line 277: Why don't we call here the "readObjectToChain" so we don't need the for loop at "readObjectsToChain", I know it will be hidding recursiveness but the level or recursion here is super low.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+	key := secret.Data[corev1.TLSPrivateKeyKey]
+	cert := secret.Data[corev1.TLSCertKey]
+	if key == nil || cert == nil {
+		return
+	}
+
+	name := serviceHostname(object.key.NamespacedName.Name, object.key.NamespacedName.Namespace)
+
+	certificateChain.CertificatesIssued[name].KeyPEM = key
+	certificateChain.CertificatesIssued[name].CertPEM = cert
```
qinqon, line 329: I suppose we don't want to do the decoding into x506.Certificate becase we want to decouple between PEM (which is an standard way of represent this info) and the golang structs, but I am not sure that we need this decoupling.

If we do the decoding here the chain constructors will be massively simplified.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+}
+
+func mapCASecretToChain(object *keyedObject, certificateChain *chain.CertificateChainData) {
+	secret := object.kobject.(*corev1.Secret)
+	key := secret.Data[CAPrivateKeyKey]
+	cert := secret.Data[CACertKey]
+	if key == nil || cert == nil {
```
qinqon, line 336: ditto

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+		return
+	}
+	mapServiceSecretFromChain(object, certificateChain)
+}
+
+func mapServiceSecretToChain(object *keyedObject, certificateChain *chain.CertificateChainData) {
+	secret := object.kobject.(*corev1.Secret)
+	key := secret.Data[corev1.TLSPrivateKeyKey]
+	cert := secret.Data[corev1.TLSCertKey]
+	if key == nil || cert == nil {
```
qinqon, line 322: We have to also chek the []byte length if we don't do so we can end with empty key/cert

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+func mapSecretFromChain(object *keyedObject, certificateChain *chain.CertificateChainData) {
+	if object.key.NamespacedName.String() == certificateChain.CA.Name {
+		mapCASecretFromChain(object, certificateChain)
+		return
+	}
+	mapServiceSecretFromChain(object, certificateChain)
+}
+
+func mapServiceSecretToChain(object *keyedObject, certificateChain *chain.CertificateChainData) {
+	secret := object.kobject.(*corev1.Secret)
```
qinqon, line 319: We are exposing ourselves to a panic, I would prefer to return and error and do a controlled downcasting

```golang
secret, ok := object.kobject.(*corev1.Secret)
if !ok {
    return ....
}

For this we have to be able to return an error here.
```

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+	}
+
+	name := serviceHostname(object.key.NamespacedName.Name, object.key.NamespacedName.Namespace)
+
+	certificateChain.CertificatesIssued[name].KeyPEM = key
+	certificateChain.CertificatesIssued[name].CertPEM = cert
+}
+
+func mapCASecretToChain(object *keyedObject, certificateChain *chain.CertificateChainData) {
+	secret := object.kobject.(*corev1.Secret)
```
qinqon, line 333: ditto

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+	cert := secret.Data[CACertKey]
+	if key == nil || cert == nil {
+		return
+	}
+	certificateChain.CA.KeyPEM = key
+	certificateChain.CA.CertPEM = cert
+}
+
+func mapServiceSecretFromChain(object *keyedObject, certificateChain *chain.CertificateChainData) {
+	secret := object.kobject.(*corev1.Secret)
```
qinqon, line 344: ditto

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+	name := serviceHostname(object.key.NamespacedName.Name, object.key.NamespacedName.Namespace)
+	bundle := certificateChain.CertificatesIssued[name]
```
qinqon, line 346: Can we rename "bundle" ? the bundle is the PEM that contains more than one certificate so that's why is named bundle, in this case this is a KeyPair or KeyCert, maybe we can name this serviceKeyCert since we are using a serviceHostName to retrieve it.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+	if bundle == nil {
+		return
 	}
-	return webhook, nil
+	if secret.Data == nil {
+		secret.Data = map[string][]byte{}
+	}
+	secret.Type = corev1.SecretTypeTLS
+	secret.Data[corev1.TLSPrivateKeyKey] = bundle.KeyPEM
+	secret.Data[corev1.TLSCertKey] = bundle.CertPEM
```
qinqon, line 355: Again doing the encoding here make sense, but all depends on how important the decoupling is how feasible is that in the future we are going to use something different than the golang api structs.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get %s webhook configuration %s", m.webhookType, m.webhookName)
+func mapCASecretFromChain(object *keyedObject, certificateChain *chain.CertificateChainData) {
+	secret := object.kobject.(*corev1.Secret)
+	secret.Type = corev1.SecretTypeOpaque
+	if secret.Data == nil {
+		secret.Data = map[string][]byte{}
 	}
+	secret.Data[CAPrivateKeyKey] = certificateChain.CA.KeyPEM
+	secret.Data[CACertKey] = certificateChain.CA.CertPEM
```
qinqon, line 365: ditto

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
-	for _, clientConfig := range m.clientConfigList(configuration) {
+func mutatingWebhookConfig(webhook client.Object) *admissionregistrationv1.MutatingWebhookConfiguration {
+	return webhook.(*admissionregistrationv1.MutatingWebhookConfiguration)
+}
 
-		service := types.NamespacedName{}
-		hostnames := []string{}
+func validatingWebhookConfig(webhook client.Object) *admissionregistrationv1.ValidatingWebhookConfiguration {
+	return webhook.(*admissionregistrationv1.ValidatingWebhookConfiguration)
+}
```
qinqon, line 421: Let's remove this helpers there are used in just one place and open the door to future changes using them.

pkg/certificate/configuration.go
```diff
@@ -2,169 +2,453 @@ package certificate
...
+//
+// The WebhookClientConfig type is share between mutating or validating so we can have a common function
+// that uses the interface runtime.Object and do some type checking to access it [1].
+//
+// [1] https://godoc.org/k8s.io/kubernetes/pkg/apis/admissionregistration#WebhookClientConfig
+func clientConfigMap(webhook client.Object) map[string]*admissionregistrationv1.WebhookClientConfig {
+	clientConfigMap := map[string]*admissionregistrationv1.WebhookClientConfig{}
+	switch webhook.(type) {
+	case *admissionregistrationv1.MutatingWebhookConfiguration:
+		mutatingWebhookConfig := mutatingWebhookConfig(webhook)
```
qinqon, line 433: do the casting directly here is the only place done and it's safe and now it was like this in the legacy code but there the "mutatingWebhookConfig(" function was call more than once.

qinqon (changes requested):
First pass on configuration.go, all the comments are smaller things in general looks good. I am missing the Maps renaming from comments like https://github.com/qinqon/kube-admission-webhook/pull/47#discussion_r612308733

PR #48, opened 2021-04-27
Prepend certs instead of append

qinqon:
To implement cert overlap multiple pub keys has to life at the cert
bundle currectly this is implemented appending new certs but this is not
ok from TLS perspective since the first public key or "leaf" has to
match with the private key [1].

[1] https://github.com/golang/go/blob/master/src/crypto/tls/tls.go#L292-L294

Test PR on kubernetes-nmstate https://github.com/nmstate/kubernetes-nmstate/pull/734
Test PR on kubemacpool https://github.com/k8snetworkplumbingwg/kubemacpool/pull/289


RamLavi (approved):
/lgtm

PR #51, opened 2021-06-09
Relationship labels, Label self deployed resources

oshoval:
In order to have relationship labels (app.kubernetes.io/*)
on all managed resources,
this commit adds an `ExtraLabels` field to the Manager options.
This way, the created secrets will be labeled.

The labels are inherited from the cert manager pod.
The pod will have those labels if deployed by OLM / HCO,
otherwise it won't and the labels won't be added to the above
resources.

It is the user responsibility to add the extra labels
to the options.

Signed-off-by: Or Shoval <oshoval@redhat.com>

### DIFF
--- a/pkg/certificate/manager_test.go
+++ b/pkg/certificate/manager_test.go
@@ -100,15 +100,35 @@ var _ = Describe("certificate manager", func() {
 		shouldFail        bool
 	}
 
-	newManager := func() *Manager {
+	genericNewManager := func(labels map[string]string) *Manager {
+		options := Options{
+			WebhookName: expectedMutatingWebhookConfiguration.ObjectMeta.Name,
+			WebhookType: MutatingWebhook, Namespace: expectedNamespace.Name,
+			CARotateInterval:   time.Hour,
+			CertRotateInterval: time.Hour,
+		}
+		if labels != nil {
+			options.ExtraLabels = labels
+		}
 
-		manager, err := NewManager(cli, Options{WebhookName: expectedMutatingWebhookConfiguration.ObjectMeta.Name, WebhookType: MutatingWebhook, Namespace: expectedNamespace.Name, CARotateInterval: time.Hour, CertRotateInterval: time.Hour})
-		ExpectWithOffset(1, err).To(Succeed(), "should success creating certificate manager")
+		manager, err := NewManager(cli, options)
+		ExpectWithOffset(2, err).To(Succeed(), "should success creating certificate manager")
 		err = manager.rotateAll()
-		ExpectWithOffset(1, err).To(Succeed(), "should success rotating certs")
+		ExpectWithOffset(2, err).To(Succeed(), "should success rotating certs")
 
 		return manager
 	}
+
+	newManager := func() *Manager {
+		return genericNewManager(nil)
+	}
+
+	const expectedLabelKey = "foo"
+	const expectedLabelValue = "bar"
+	newManagerWithLabels := func() *Manager {
+		return genericNewManager(map[string]string{expectedLabelKey: expectedLabelValue})
+	}
+
 	loadServiceSecret := func(manager *Manager) corev1.Secret {
 		secretKey := types.NamespacedName{
 			Namespace: expectedSecret.ObjectMeta.Namespace,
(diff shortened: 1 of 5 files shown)


oshoval:
/assign @qinqon 
Tested manually on k8s-nmstate (vendored changes, and did what PR describes)

pkg/certificate/secret.go
```diff
@@ -110,6 +111,7 @@ func (m *Manager) applySecret(secretKey types.NamespacedName, secretType corev1.
 						Name:        secretKey.Name,
 						Namespace:   secretKey.Namespace,
 						Annotations: map[string]string{},
+						Labels:      names.RelationshipLabels,
```
oshoval, line 114: @qinqon 
since you asked on https://github.com/nmstate/kubernetes-nmstate/pull/770
to remove the global var and just always calculate it

do we want to do here the same ?
always read the env vars?
or we want to change the constructor to have another parameter,
but then we will need to save that parameter in some global var

qinqon: Don't use a global pass it over the functions until it reaches this, is better to be explicit with dependencies.

oshoval: better to just run the function isn't it ?
won't it be too much sending in the call stack ?

oshoval: ah its will be part of Manager, and the function is a method of manager so we are good

oshoval: Just note that we will need to add logic of https://github.com/nmstate/kubernetes-nmstate/pull/770
also to kubemacpool because kubemacpool will also need to pass the argument

oshoval: Done, just need to think on the tests if any

RamLavi (commented):
question. 
Since KAW is loaded in a very early stage of the product's (knmstate for example) life, is it possible that the labels are not yet set?

oshoval:
> question.
> Since KAW is loaded in a very early stage of the product's (knmstate for example) life, is it possible that the labels are not yet set?

I don't think so, 
because CNAO will create the deployment /daemonset with the labels, and the pod that is created there will have them as env var upon its start

oshoval:
rebase

oshoval:
Added unit tests,
CA secret deletion was missing in `deleteResources`
we might want / need to add Eventually to check that the CA secret is deleted, else it might flakes randomly

RamLavi (approved):
/lgtm

oshoval:
Thanks Ram
@qinqon can you take a look please?

pkg/certificate/manager_test.go
```diff
@@ -157,6 +182,32 @@ var _ = Describe("certificate manager", func() {
...
+		Entry("when manager created with labels, should have labels", verifyLabelsTestCase{
+			createManager:    newManagerWithLabels,
+			shouldHaveLabels: true,
+		}),
+		Entry("when manager created without labels, should have no labels", verifyLabelsTestCase{
+			createManager:    newManager,
+			shouldHaveLabels: false,
+		}),
+	)
+
```
qinqon, line 210: There is a pair of tests here it does not worth it a DescriptionTable half of the function is different per test, please create individual Specs

```golang
Expect(obtainedCASecret.GetLabels()).To(HaveKeyWithValue(expectedLabelKey, expectedLabelValue)
```

and

```golang
Expect(obtainedCASecret.GetLabels()).To(BeEmpty())
```

qinqon (changes requested):
Looks good just some test reorderning.

oshoval:
Addressed comments

pkg/certificate/manager_test.go
```diff
@@ -157,6 +177,30 @@ var _ = Describe("certificate manager", func() {
...
+			createResources()
+			defer deleteResources()
+			manager := newManagerWithLabels()
+			obtainedCASecret := loadCASecret(manager)
+			Expect(obtainedCASecret.GetLabels()).To(HaveKeyWithValue(expectedLabelKey, expectedLabelValue))
+			obtainedSecret := loadServiceSecret(manager)
+			Expect(obtainedSecret.GetLabels()).To(HaveKeyWithValue(expectedLabelKey, expectedLabelValue))
+		})
+	})
+
```
qinqon, line 203: Can you group it like this? :

```golang
Context("when secrets are created", func(){
    BeforeEach(func(){
        createResources()
    })
    AfterEach(func(){
        deleteResources()
    })
    Context("with extra labels option", func() {
        var manager
        BeforeEach(func() {
            manager = newManagerWithLabels
        })
        It("should have extra labels", func()...
    })
   Context("without extra labels option", func(){
        var manager
        BeforeEach(func() {
            manager = newManager
        })
        It("should not have extra labels", func()...
   }}
}
```

qinqon (changes requested):
nit: A little more structured test

oshoval:
Addressed comments

qinqon (approved):

PR #54, opened 2021-10-05
certs, Add Certs list size limit

RamLavi:
Currently certs are appended without limit. This can
cause a problem when applying the caBundle/certs on the
mutatingwebhookconfiguration/secret, appropriately - due
to etcd's POST size limit.
Limited the amount of certs to 100.


phoracek:
@RamLavi is this (chain of 100 certs) something that can happen with our rotation defaults?

RamLavi:
> @RamLavi is this (chain of 100 certs) something that can happen with our rotation defaults?

if the rotation cycle is short,  we can accumulate certs on the caBundle quite quickly, since we are currently appending the certs (always on caBundle, on secrets it is less of a problem since we flush the certs upon overlap). 
It was observed that after ~700 rotations the mutatingWebhookConfiguration becomes too big to apply to etcd due to how large the caBundle becomes.

This limit can also be reached if there are alot of retry on conflict on caBundle apply.

phoracek:
On Wed, 6 Oct 2021, 10:25 RamLavi, ***@***.***> wrote:

> @RamLavi <https://github.com/RamLavi> is this (chain of 100 certs)
> something that can happen with our rotation defaults?
>
> if the rotation cycle is short, we can accumulate certs on the caBundle
> quite quickly, since we are currently appending the certs (always on
> caBundle, on secrets it is less of a problem since we flush the certs upon
> overlap).
> It was observed that after ~700 rotations the mutatingWebhookConfiguration
> becomes too big to apply to etcd due to how large the caBundle becomes.
>

Do I understand this correctly that the same problem will happen on a
system that rotates every day after 2 years of use? Or do we have a cleanup
option in place. I'm trying to see if we need to backport this and how far.

This limit can also be reached if there are alot of retry on conflict on

RamLavi:
>Do I understand this correctly that the same problem will happen on a
system that rotates every day after 2 years of use? Or do we have a cleanup
option in place. I'm trying to see if we need to backport this and how far.

As a temp cleanup option we have the existing workaround [force-cert-retation.sh](https://github.com/qinqon/kube-admission-webhook/blob/master/hack/force-cert-retation.sh) which removed the caBundles from the mutatinWebhokConfiguration and fixed the problem on the machine. 

As for the 2 years, I believe you are right, but note that in our machine it happened after a few days. Remembering that knmstate has this exact certs pod where the issue did not occur, I suspect this happened there due to some tests on KMP that caused the certs to get rotated sooner.

pkg/certificate/triple/pem.go
```diff
@@ -200,9 +202,24 @@ func AddCertToPEM(cert *x509.Certificate, pemCerts []byte) ([]byte, error) {
 	// Prepend cert since it's what TLS expects [1]
 	// [1] https://github.com/golang/go/blob/master/src/crypto/tls/tls.go#L292-L294
 	certs = append([]*x509.Certificate{cert}, certs...)
+
+	if len(certs) > maxListSize {
+		certs = removeOldestCerts(certs, maxListSize)
+	}
```
qinqon, line 208: You can directly do

```golang
certs = certs[:maxListSize]
```

qinqon (changes requested):
We don't need to loop.

PR #60, opened 2021-12-01
certificate: Reconcile only created secrets

qinqon:
If two cert-managers are running at a cluster they will each other
secrets since the kube-admission-webhook annotation is there. This
change filter the secrets and only reconcile the ones with the CA cert
name from webhook config and the services owned by them.

### DIFF
--- a/pkg/certificate/controller.go
+++ b/pkg/certificate/controller.go
@@ -43,19 +43,48 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return object.GetName() == m.webhookName
 	}
 
+	isCASecret := func(object client.Object) bool {
+		return object.GetName() == m.caSecretKey().Name
+	}
+
+	isServiceSecret := func(object client.Object) bool {
+		webhookConf, err := m.readyWebhookConfiguration()
+		if err != nil {
+			m.log.Info(fmt.Sprintf("failed checking if it's a generated secret: failed getting webhook configuration: %v", err))
+			return false
+		}
+
+		services, err := m.getServicesFromConfiguration(webhookConf)
+		if err != nil {
+			m.log.Info(fmt.Sprintf("failed checking if it's a generated secret: failed getting webhook configuration services: %v", err))
+			return false
+		}
+
+		for service, _ := range services {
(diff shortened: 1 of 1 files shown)


pkg/certificate/controller.go
```diff
@@ -43,19 +43,43 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return object.GetName() == m.webhookName
 	}
 
+	isGeneratedSecret := func(object client.Object) bool {
+		if object.GetName() == m.caSecretKey().Name {
+			return true
+		}
```
RamLavi, line 49: can we move this to a specific function `isCASecretRelated`

RamLavi, line 49: can we move [this](https://github.com/qinqon/kube-admission-webhook/pull/60/files#diff-477009fdb255fbbfc90de2def9ee37495b402bf4e99058b8fdf409341c11f066R47-R49) to a specific function `isCASecretRelated` (or any better name you can find)

pkg/certificate/controller.go
```diff
@@ -43,19 +43,43 @@ func (m *Manager) add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		if err != nil {
+			m.log.Info(fmt.Sprintf("failed checking if it's a generated secret: failed getting webhook configuration services: %v", err))
+			return false
+		}
+
+		for service, _ := range services {
+			if object.GetName() == service.Name {
+				return true
+			}
+		}
```
RamLavi, line 67: can we move [this](https://github.com/qinqon/kube-admission-webhook/pull/60/files#diff-477009fdb255fbbfc90de2def9ee37495b402bf4e99058b8fdf409341c11f066R51-R67) to a specific function `isServiceSecretRelated` (or any better name you can find)

This way `isGeneratedSecret` will be more readable to a stranger:
```
return isCASecretRelated(object) || isServiceSecretRelated(object)
```

RamLavi (changes requested):
looks good. 
Suggested to move to human readable functions for better readability, but not a must.

RamLavi (approved):
/lgtm

PR #61, opened 2022-03-16
server: Remove webhook server

qinqon:
The webhook server that start a cert-manager inside of it is no longer
in use and introduces a chicken egg issue since the cert/key has to
exist for the pod to be start up, apart from the need for a leader
election mechanism and the latency on the pod projecting its own changed
cert/key. This commit remove the server.


rhrazdil (approved):
/lgtm
/approve

PR #65, opened 2022-03-16
webhook: Implement controller-runtime webhook server with CipherSuites knob

qinqon:
The current controller-runtime allow to configure TLS min version but it
does not allow to configure the cipher suites. This change copy the
server from the controller-runtime library and add a CipherSuites knob to configure the TLS handshake cipher algorithms.

controller-runtime https server [does not expose](https://github.com/kubernetes-sigs/controller-runtime/issues/852#issuecomment-1062538456) the `tls.Config`, this PR has copied the server there to be able to modify tls.Config.

### DIFF
--- a/pkg/webhook/internal/metrics/metrics.go
+++ b/pkg/webhook/internal/metrics/metrics.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2022 The Kube Admission Webhook Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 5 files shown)


pkg/webhook/internal/metrics/metrics.go
```diff
@@ -0,0 +1,85 @@
+/*
+Copyright 2022 The Kube AdmissionWebhook Authors.
```
rhrazdil, line 2: missing space in AdmissionWebhook

rhrazdil (commented):
One super small nit

rhrazdil (approved):

PR #71, opened 2022-04-05
actions: Remove GITHUB_TOKEN from arnested/go-version-action 

arnested:
Hi,

I'm the author of [arnested/go-version-action](https://github.com/marketplace/actions/go-version-action).

The action doesn't need a GITHUB_TOKEN anymore.

Instead of getting the Go releases from git tags using GitHub's API (and thus needing the token to avoid being rate limited) it pulls the versions from https://go.dev/dl/?mode=json&include=all.


qinqon:
Hey thanks for this!

qinqon (approved):

PR #73, opened 2023-09-07
Remove webhook server

qinqon:
Now that controller-runtime is able to configure CipherSuites we no longer need the webhook server. This change remove the custom server but keeps the pair of functions to map tls versions and cipher suites, it also bumps controller-runtime to 0.13.1 wich is the minimal that allow configuring cipher suites.


RamLavi (approved):
/lgtm

PR #80, opened 2025-11-23
logs: Add levels to logs

oshoval:
In order to reduce log noise, add level to repeating Info logs. 
Those won't be printed in production by default now.

### DIFF
--- a/pkg/certificate/cleanup.go
+++ b/pkg/certificate/cleanup.go
@@ -73,7 +73,7 @@ func (m *Manager) earliestElapsedForCleanup(log logr.Logger, certificates []*x50
 	deadline := m.earliestCleanupDeadlineForCerts(certificates)
 	now := m.now()
 	elapsedForCleanup := deadline.Sub(now)
-	log.Info(fmt.Sprintf("{now: %s, deadline: %s, elapsedForCleanup: %s}", now, deadline, elapsedForCleanup))
+	log.V(1).Info(fmt.Sprintf("{now: %s, deadline: %s, elapsedForCleanup: %s}", now, deadline, elapsedForCleanup))
 	return elapsedForCleanup, nil
 }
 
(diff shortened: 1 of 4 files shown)


oshoval:
cc @qinqon @RamLavi

RamLavi (commented):
Hey @oshoval 
- Is the log level configurable? if not, can you add this option? As a debugger trying to find a bug I would want to be able to get more information when needed..
- can you paste how the new logs will look like? it will make it easier to understand if we have a good log picture from the current info logging.

oshoval:
> Is the log level configurable? if not, can you add this option? As a debugger trying to find a bug I would want to be able to get more information when needed..

By changing the existing `--v=debug` / `--v=production` toggles the level
`ctrl.SetLogger(zap.New(zap.UseDevMode(logType != "production")))`

https://github.com/k8snetworkplumbingwg/kubemacpool/blob/6ea040ae7f12e8c03e88e3ef0e672cf1d6de7a2e/vendor/sigs.k8s.io/controller-runtime/pkg/log/zap/zap.go#L171

oshoval:
its pretty pain to test it before it is merged (created a setup for that),
and the whole change might doesn't worth it (the worse there is 1 per hour in average, and there aren't many)

i suggest that alternatively to dump how the log looks like
if @qinqon can look please on the logs that had changed, and say if you think you need them on debug / production level ?

file 2 here might help
https://gist.github.com/oshoval/bcf4432dae430e35c8261ab2b8dbe7df

worst case we can just close this PR

oshoval:
lets close it, unless @qinqon suggest otherwise
i don't think it worth it for a log that appear 1 / hour at most (the rest appear even less)

RamLavi:
@oshoval if you can provide a quay image I could check it on kubemacpool locally

oshoval:
thanks, i dont think its needed, (looked on OCP cluster of QE, that ran 14 days, it didn't spam almost)

steps that i did to cluster-sync KMP with it locally, just in case we will need it again

1. git clone the repo under KMP `kube-admission-webhook`
2. add to go.mod
```
replace github.com/qinqon/kube-admission-webhook => ./kube-admission-webhook
```
3. in `build/Dockerfile`
add `COPY kube-admission-webhook ./kube-admission-webhook` before `RUN go mod download`
4. run `make vendor` (important after every manual change to the cloned folder)

